PR #23, opened 2021-06-29
Flux monitor smoke test

skudasov:


kalverra (changes requested):
Overall good stuff, one nitpick, and a couple of open questions that I'll leave for @jleeh to answer definitively.

contracts/contract_models.go
```diff
@@ -29,23 +29,36 @@ type FluxAggregatorData struct {
...
-		client.BlockchainWallet,
-		[]common.Address,
-		[]common.Address,
-		[]common.Address,
-		uint32,
-		uint32,
-		uint32,
-	) error
+	AvailableFunds(ctx context.Context) (*big.Int, error)
+	AllocatedFunds(ctx context.Context) (*big.Int, error)
```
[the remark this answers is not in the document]
skudasov: Ok, got it. Maybe we can skip some checks but have implementation available when we need them, or remove implementations too?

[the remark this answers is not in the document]
jleeh: I'd remove them unless we need them or it'll result in us needing to create boiler plate implementations that just adds effort.

client/chainlink.go
```diff
@@ -62,22 +64,43 @@ func NewChainlink(c *ChainlinkConfig, httpClient *http.Client) (Chainlink, error
 	return cl, cl.SetSessionCookie()
 }
 
+func FundTemplateNodes(blockchainClient BlockchainClient, wallets BlockchainWallets, nodes []Chainlink, ethAmount int64, linkAmount int64) error {
```
jleeh, line 67: The client is specifically for API interactions with the Chainlink node, it needs to have no dependencies on anything blockchain related. These sorts of actions that combine clients & blockchain should go into another package.

client/chainlink.go
```diff
@@ -62,22 +64,43 @@ func NewChainlink(c *ChainlinkConfig, httpClient *http.Client) (Chainlink, error
...
+			return err
+		}
+	}
+	return nil
+}
+
 // ConnectToTemplateNodes assumes that 5 template nodes are running locally, check out a quick setup for that here:
 // https://github.com/smartcontractkit/chainlink-node-compose
-func ConnectToTemplateNodes() ([]Chainlink, error) {
+func ConnectToTemplateNodes() ([]Chainlink, []common.Address, error) {
```
jleeh, line 89: This is a bad shortcut to me which'll be removed in the future when we implement k8s deployments. It should be taken out of this client.

contracts/contracts_flux_test.go
```diff
@@ -0,0 +1,172 @@
...
+				Name:              "flux_monitor",
+				ContractAddress:   fluxInstance.Address(),
+				PollTimerPeriod:   15 * time.Second, // min 15s
+				PollTimerDisabled: false,
+				ObservationSource: ObservationSourceSpec(adapter.InsideDockerAddr + "/variable"),
+			}
+			_, err = n.CreateJob(fluxSpec)
+			Expect(err).ShouldNot(HaveOccurred())
+		}
+		time.Sleep(5 * time.Second)
```
jleeh, line 72: No sleeps in tests

contracts/helper.go
```diff
@@ -0,0 +1,59 @@
...
+	observationSourceTmpl = `fetch    [type=http method=POST url="%s" requestData="{}"];
+			parse    [type=jsonparse path="data,result"];    
+			fetch -> parse;`
+)
+
+func ObservationSourceSpec(url string) string {
+	return fmt.Sprintf(observationSourceTmpl, url)
+}
+
+func DefaultLocalSetup(initFunc client.BlockchainNetworkInit) (*DefaultSuiteSetup, error) {
```
jleeh, line 27: Nice function, but this should go into its own package as it's not contract related.

contracts/helper.go
```diff
@@ -0,0 +1,59 @@
...
+	Link     LinkToken
+}
+
+var (
+	observationSourceTmpl = `fetch    [type=http method=POST url="%s" requestData="{}"];
+			parse    [type=jsonparse path="data,result"];    
+			fetch -> parse;`
+)
+
+func ObservationSourceSpec(url string) string {
```
jleeh, line 23: This doesn't seem to make sense being in here?

jleeh (changes requested):

contracts/ethereum_contracts.go
```diff
@@ -23,11 +25,152 @@ type EthereumFluxAggregator struct {
...
+}
+
+// AwaitNextRound awaits for the next round to happen, see config.yaml for default values
+func (f *EthereumFluxAggregator) AwaitNextRound(ctx context.Context) error {
+	lr, err := f.LatestRound(ctx)
+	if err != nil {
+		return err
+	}
+	log.Info().Int64("round", lr.Int64()).Msg("awaiting next round after")
+	if err := retry.Do(func() error {
```
jleeh, line 129: Love the retry mechanism, I'm just wondering if this logic would be better to be converted to a log subscription and timeout so it doesn't have to poll?

jleeh (changes requested):
Single comment, but great first PR @skudasov

contracts/ethereum_contracts.go
skudasov: Tried log subscription, I think polling is ok in that case because we need to know when the round is finalized, not started.
Will do event awaiting and logging in a separate PR.

jleeh (approved):

kalverra (approved):

PR #25, opened 2021-07-07
Volume tests

skudasov:
Volume test for FM, deploys 100 FM jobs across nodes, triggers new round, and check each round completion time
Works only on Geth dev network with our k8s cluster
```
NETWORK="ethereum_geth_volume" make test_performance
```


suite/contracts/contracts_volume_test.go
```diff
@@ -0,0 +1,155 @@
...
+		// all contracts/nodes/jobs are setup at that point, triggering new round,
+		// waiting for all contracts to complete one round and get metrics
+		err := adapter.TriggerValueChange(iteration)
+		if err != nil {
+			log.Fatal().Err(err)
+		}
+		// TODO: check all contracts have one round completed, options are:
+		// TODO: 1. poll prom metrics for last seen round, ensure all nodes have complete next round and there is no errors
+		// TODO: 2. override contract submit method, when round is complete call another collector contract
+		// TODO: when all nodes have complete one round listen to RoundComplete event
```
skudasov, line 143: For #2 I mean that metrics:
```
# HELP flux_monitor_reported_round Flux monitor's last reported round
# TYPE flux_monitor_reported_round gauge
flux_monitor_reported_round{job_spec_id="1"} 10
flux_monitor_reported_round{job_spec_id="10"} 12
flux_monitor_reported_round{job_spec_id="2"} 9
```

client/ethereum.go
```diff
@@ -265,6 +304,28 @@ func (e *EthereumClient) WaitForTransaction(transactionHash common.Hash) error {
...
+				Int("Total Confirmations", confirmations)
+			isConfirmed, err := e.isTxConfirmed(transactionHash)
+			if err != nil {
+				return err
+			} else if !isConfirmed {
+				confirmationLog.Msg("Transaction still pending, waiting for confirmation")
+				continue
+			}
+			confirmations++
+			confirmationLog.Msg("Transaction confirmed, waiting on confirmations")
```
skudasov, line 328: Just polling, I think it's out of scope, but no new header was received in the previous case branch, so we were always waiting for timeout.
Make it for being able to run x100 batch deployments, works fine.

scripts/local_prometheus/README.md
```diff
@@ -0,0 +1,15 @@
...
+```
+Visit UI
+```
+http://localhost:9090 - Prometheus UI
+http://localhost:6711/metrics - first Chainlink node metrics
+http://localhost:9100/metrics - node exporter metrics
+```
+Default docker-compose targets are in `prometheus.yml`
+
+Also watch default config for Prometheus client in `./config/config.yaml`
```
skudasov, line 15: Prometheus might be not always available in cluster, just for local debug.

suite/contracts/contracts_volume_test.go
skudasov: For #2 we can override submit method and measure time from round start to round end, also call another contract (volume_collector. sol) when submissions are enough, collector emits an event when all nodes have complete one round

skudasov: Anyway, I did both, gonna check for prom metrics and ContractData across all deployed aggregators.

suite/volume/contracts_volume_setup.go
```diff
@@ -0,0 +1,193 @@
...
+		}
+		roundValsArr = append(roundValsArr, roundVals{
+			RoundID: r.LatestRoundData.RoundId.Int64(),
+			Val:     r.LatestRoundData.Answer.Int64(),
+		})
+	}
+	log.Debug().Interface("rounds_on_chain", roundValsArr).Msg("last rounds on chain")
+}
+
+func (vt *VolumeFluxTest) checkRoundFinishedOnChain(roundID int, newVal int) bool {
```
jleeh, line 175: Anything returning a bool should start with `is`, so `isRoundFinishedOnChain`

suite/volume/hardhat.config.js
```diff
@@ -0,0 +1,18 @@
...
+module.exports = {
+    solidity: "0.8.6",
+    networks: {
+        // blockGasLimit and interval may vary for heavy concurrent test deployments
+        hardhat: {
+            // x10 of default block gas limit
+            blockGasLimit: 124500000,
+            mining: {
+                auto: false,
+                interval: 1000
```
jleeh, line 13: I'd remove mining intervals from load testing as it'll skew results.

client/ethereum.go
```diff
@@ -237,11 +272,15 @@ func (e *EthereumClient) WaitForTransaction(transactionHash common.Hash) error {
...
 
+	ticker := time.NewTicker(200 * time.Millisecond)
+	defer ticker.Stop()
+
 	for {
 		select {
 		case err := <-subscription.Err():
 			return err
 		case header := <-headerChannel:
+			// FIXME: investigate why there is no new header with hardhat mining
```
jleeh, line 283: Just make sure this is fixed prior to merge.

config/config.yml
```diff
@@ -54,3 +55,8 @@ networks:
 retry:
     attempts: 20
     linear_delay: 1s
+prometheus:
```
jleeh, line 58: Prometheus shouldn't be handled separately and in this case locally, it should be deployed within the environment. and be specific to that environment. It'll differ between compose/k8s.

scripts/local_prometheus/prometheus.yml
```diff
@@ -0,0 +1,7 @@
+global:
+  scrape_interval: 5s
+scrape_configs:
+  - job_name: 'local_scrape'
+    scrape_interval: 1s
+    static_configs:
+      - targets: ['0.0.0.0:6711', '0.0.0.0:6722', '0.0.0.0:6733', '0.0.0.0:6744', '0.0.0.0:6755', '0.0.0.0:9100']
```
jleeh, line 7: Should be handled within docker-compose and using service discovery, not hardcoded ports.

scripts/local_prometheus/local_node_exporter.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
```
jleeh, line 1: When handled in the env these scripts can be deleted.

tools/prometheus.go
```diff
@@ -0,0 +1,213 @@
...
+	API v1.API
+	Cfg *config.PrometheusClientConfig
+}
+
+func NewPrometheusClient(cfg *config.PrometheusClientConfig) *PromChecker {
+	client, err := api.NewClient(api.Config{
+		Address: cfg.Url,
+	})
+	if err != nil {
+		log.Fatal().Err(err)
```
jleeh, line 44: No fatals, anywhere

jleeh (changes requested):

scripts/local_prometheus/local_node_exporter.sh
skudasov: Move them to compose here https://github.com/smartcontractkit/chainlink-node-compose/pull/7

client/ethereum.go
skudasov: Tried to refactor it using background goroutine and concurrent map, it seems subscription is completely not working in hardhat for now because of https://github.com/nomiclabs/hardhat/pull/1163
Also tried it with geth - it's working, maybe we can use poll for now, and add required changes when we add geth?

config/config.yml
jleeh: If this is resolved, I'd expect this config to be removed.

suite/volume/flux_volume_test.go
```diff
@@ -0,0 +1,86 @@
...
+			// all contracts/nodes/jobs are setup at that point, triggering new round,
+			// waiting for all contracts to complete one round and get metrics from runs/on-chain
+			newVal, err := ft.Adapter.TriggerValueChange(currentRound)
+			Expect(err).ShouldNot(HaveOccurred())
+
+			ctx, cancel := context.WithTimeout(context.Background(), promRoundTimeout)
+			defer cancel()
+
+			// await all nodes report round was finished
+			roundFinished, err := ft.Prom.AwaitRoundFinishedAcrossNodes(ctx, currentRound+1)
```
jleeh, line 52: We're also checking on-chain to see if the rounds completed, I wouldn't have both a prometheus & on-chain check. On chain is what we should solely depend on.

suite/volume/volume_setup.go
```diff
@@ -0,0 +1,335 @@
...
+				}
+			}
+		}
+	}
+	log.Debug().Interface("round_end_block_times", endTimesMap).Send()
+	return endTimesMap, nil
+}
+
+// getOnChainLogs get all block data for test interval
+func (vt *FluxTest) getOnChainLogs(from *big.Int, to *big.Int, roundID int) (*ethlog.HistoryResult, error) {
```
jleeh, line 314: This is ETH specific, we should have 0 ETH specific logic outside of the contract implementations.

suite/volume/volume_setup.go
```diff
@@ -0,0 +1,335 @@
...
+	for _, r := range rounds {
+		if r.LatestRoundData.RoundId.Int64() != int64(roundID) || r.LatestRoundData.Answer.Int64() != int64(newVal) {
+			return false
+		}
+	}
+	return true
+}
+
+// roundEndTimes counts submissions for round, when all required submission is found, saves last block_timestamp for contract
+func (vt *FluxTest) roundEndTimes(hr *ethlog.HistoryResult, expectedRound int, submitsRequired int) (map[string]int64, error) {
```
jleeh, line 279: ETH specific and this looks very messy.

suite/volume/volume_setup.go
```diff
@@ -0,0 +1,335 @@
...
+// JobByInstance helper struct to match job + instance ID against prom metrics
+type JobByInstance struct {
+	ID       string
+	Instance string
+}
+
+// NewFluxTest deploys AggregatorsNum flux aggregators concurrently
+func NewFluxTest(spec *FluxTestSpec) *FluxTest {
+	s, err := suite.DefaultLocalSetup(spec.InitFunc)
+	Expect(err).ShouldNot(HaveOccurred())
```
jleeh, line 71: No test errors outside of tests.

suite/volume/volume_setup.go
```diff
@@ -0,0 +1,335 @@
...
+			FluxInstancesMu:   mu,
+			NodesByHostPort:   nodesByHostPort,
+			FluxInstances:     &fluxInstances,
+			ContractToJobsMap: contractToJobsMap,
+		})
+	}
+	wg.Wait()
+	prom, err := tools.NewPrometheusClient(s.Config.Prometheus)
+	Expect(err).ShouldNot(HaveOccurred())
+	log.Debug().Interface("contracts_to_jobs", contractToJobsMap).Msg("debug data for per round metrics")
```
jleeh, line 109: I had a previous comment about logging style.

suite/volume/volume_setup.go
```diff
@@ -0,0 +1,335 @@
...
+		}
+		d.FluxInstancesMu.Lock()
+		*d.FluxInstances = append(*d.FluxInstances, fluxInstance)
+		d.FluxInstancesMu.Unlock()
+	}()
+}
+
+// runInstanceKey to record start timestamp of every run + instance
+func (vt *FluxTest) runInstanceKey(inst string, runID string) string {
+	return fmt.Sprintf("%s:%s", inst, runID)
```
jleeh, line 179: I don't understand why this is needed

tools/prometheus.go
```diff
@@ -0,0 +1,231 @@
+package tools
```
jleeh, line 1: I don't think I'm a fan of leveraging Prometheus at all for volume tests.

jleeh (changes requested):
On reviewing this the second time round, I think there's too much code, logic and complexity for what is being done here. This could be cut down, there's still ETH specific logic, bad error handling, mis-matched logging styles and just too much overall complexity.

I think this PR needs re-working to work with @kalverra's k8s branch and then we can revisit it.

suite/volume/volume_setup.go
skudasov: Because run ID's are not unique across nodes, so I cache them all in one map. Tried to use /v2/bulk_delete_runs but it doesn't work.

contracts/ethereum_contracts.go
```diff
@@ -23,6 +24,39 @@ type EthereumFluxAggregator struct {
 	address        *common.Address
 }
 
+// FilterRoundSubmissions filters rounds submissions, if there isn't enogh submissions found returns and error
+func (f *EthereumFluxAggregator) FilterRoundSubmissions(ctx context.Context, submissions int, submissionVal *big.Int, roundID int) (int64, error) {
```
jleeh, line 28: Logic is too specific within a contract function, it should be a wrapper client. Test logic should handle whether there's enough submissions etc.

suite/volume/common.go
```diff
@@ -0,0 +1,116 @@
...
+		return nil, err
+	}
+	stdDev, err := stats.StandardDeviation(data)
+	if err != nil {
+		return nil, err
+	}
+	return &PercentileReport{P99: perc99, P95: perc95, P90: perc90, P50: perc50, Max: max, Min: min, StdDev: stdDev}, nil
+}
+
+func minInt64Slice(v []int64) (m int64) {
```
jleeh, line 94: Use the built-in sort package for this and max:
https://pkg.go.dev/sort

suite/volume/flux_volume_test.go
```diff
@@ -0,0 +1,65 @@
...
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/smartcontractkit/integrations-framework/client"
+	"github.com/smartcontractkit/integrations-framework/contracts"
+	"github.com/smartcontractkit/integrations-framework/tools"
+	"math/big"
+	"time"
+)
+
+var _ = Describe("Flux monitor volume tests", func() {
```
jleeh, line 14: Test implementation is great

suite/volume/hardhat.config.js
jleeh: Actioning this?

suite/volume/volume_setup.go
```diff
@@ -0,0 +1,320 @@
...
+	}
+	clNodes, nodeAddrs, err := suite.ConnectToTemplateNodes()
+	if err != nil {
+		return nil, err
+	}
+	err = suite.FundTemplateNodes(s.Client, s.Wallets, clNodes, 9e18, 0)
+	if err != nil {
+		return nil, err
+	}
+	s.Client.(*client.EthereumClient).BorrowedNonces(true)
```
jleeh, line 79: No ETH specific calls..

suite/volume/volume_setup.go
```diff
@@ -0,0 +1,320 @@
...
+		return nil, err
+	}
+
+	fluxInstances := make([]contracts.FluxAggregator, 0)
+	nodesByHostPort := make(map[string]client.Chainlink)
+	contractToJobsMap := make(map[string][]JobByInstance)
+	mu := &sync.Mutex{}
+	g := &errgroup.Group{}
+	for i := 0; i < spec.AggregatorsNum; i++ {
+		deployFluxInstance(&FluxInstanceDeployment{
```
jleeh, line 91: Commented this before, but the goroutine call should be here, not wrapping the whole function implementation in a goroutine.

suite/volume/volume_setup.go
```diff
@@ -0,0 +1,320 @@
...
+		if err != nil {
+			return err
+		}
+		err = fluxInstance.UpdateAvailableFunds(context.Background(), d.Suite.Wallets.Default())
+		if err != nil {
+			return err
+		}
+		// set oracles and submissions
+		err = fluxInstance.SetOracles(d.Suite.Wallets.Default(),
+			contracts.SetOraclesOptions{
```
jleeh, line 149: Name of this struct implies that it's generic to all contract types. Should be named `FluxAggregatorOracleOptions`.

suite/volume/volume_setup.go
```diff
@@ -0,0 +1,320 @@
...
+			for _, j := range jobs {
+				// get node for a job
+				node := vt.NodesByHostPort[j.Instance]
+				runs, err := node.ReadRunsByJob(j.ID)
+				if err != nil {
+					return err
+				}
+				runsStartTimes := make([]int64, 0)
+				for _, r := range runs.Data {
+					runsStartTimes = append(runsStartTimes, r.Attributes.CreatedAt.UnixNano()/1e6) // milliseconds
```
jleeh, line 209: Just store the timestamp and then do a subtraction of the start time away from the end time which gives you a `time.Duration` that has built-in functions for seconds/ms/minutes etc.

suite/volume/volume_setup.go
```diff
@@ -0,0 +1,320 @@
...
+
+// roundsStartTimes gets run start time for every contract
+func (vt *FluxTest) roundsStartTimes() (map[string]float64, error) {
+	mu := &sync.Mutex{}
+	g := &errgroup.Group{}
+	contactsStartTimes := make(map[string]float64)
+	for contractAddr, jobs := range vt.ContractsToJobsMap {
+		contractAddr := contractAddr
+		jobs := jobs
+		g.Go(func() error {
```
jleeh, line 198: The routine defined in this function should be separated into it's own function, as it'll be useful elsewhere. I think we need an `actions` package that has more complex logic that's re-usable.

suite/volume/volume_setup.go
```diff
@@ -0,0 +1,320 @@
...
+}
+
+func (vt *FluxTest) getAllRoundsSubmissionsEvents(ctx context.Context, roundID int, submissions int, submissionVal *big.Int) (map[string]float64, error) {
+	g, gctx := errgroup.WithContext(ctx)
+
+	mu := &sync.Mutex{}
+	endTimes := make(map[string]float64)
+	for _, fi := range *vt.FluxInstances {
+		fi := fi
+		g.Go(func() error {
```
jleeh, line 236: Same with this function. An action called `getRoundCompleteTimestamp`.

suite/volume/volume_setup.go
```diff
@@ -0,0 +1,320 @@
...
+	startTimes, err := vt.roundsStartTimes()
+	if err != nil {
+		return err
+	}
+	endTimes, err := vt.getAllRoundsSubmissionsEvents(context.Background(), roundID, submissions, submissionVal)
+	if err != nil {
+		return err
+	}
+	for contract := range startTimes {
+		duration := endTimes[contract] - startTimes[contract]
```
jleeh, line 265: As mentioned this'd be much cleaner if it was subtracting of timestamps resulting in a `time.Duration`.

suite/volume/volume_setup.go
```diff
@@ -0,0 +1,320 @@
...
+	return true, nil
+}
+
+func (vt *FluxTest) debugRoundTuple(rounds []*contracts.FluxAggregatorData) {
+	roundValsArr := make([]roundVals, 0)
+	for _, r := range rounds {
+		if r == nil {
+			continue
+		}
+		roundValsArr = append(roundValsArr, roundVals{
```
jleeh, line 314: Do we really need a struct for this? Can't it just output the round ID & val without creating a struct?

tools/prometheus.go
```diff
@@ -0,0 +1,126 @@
...
+	"time"
+)
+
+const (
+	QueryMemoryUsage               = `100 * (1 - ((avg_over_time(node_memory_MemFree_bytes[%s]) + avg_over_time(node_memory_Cached_bytes[%s]) + avg_over_time(node_memory_Buffers_bytes[%s])) / avg_over_time(node_memory_MemTotal_bytes[%s])))`
+	QueryNodesLastReportedRoundsTs = `timestamp(flux_monitor_reported_round{job="%s",job_spec_id=~".*"})`
+	QueryAllCPUBusyPercentage      = `100 - (avg by (instance) (irate(node_cpu_seconds_total{job="%s",mode="idle"}[%s])) * 100)`
+)
+
+type FluxRoundMetrics struct {
```
jleeh, line 20: Delete these Flux structs since they're not used

tools/prometheus.go
```diff
@@ -0,0 +1,126 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return &PromChecker{
+		API: v1.NewAPI(client),
+		Cfg: cfg,
+	}, nil
+}
+
+func (p *PromChecker) NodesLastReportedRoundsTs() (model.Vector, error) {
```
jleeh, line 49: Not used func

tools/prometheus.go
```diff
@@ -0,0 +1,126 @@
...
+func (p *PromChecker) ResourcesSummary() (*FluxRoundSummary, error) {
+	cpu, err := p.CPUBusyPercentage()
+	if err != nil {
+		return nil, err
+	}
+	mem, err := p.MemoryUsage()
+	if err != nil {
+		return nil, err
+	}
+	return &FluxRoundSummary{
```
jleeh, line 122: No need to wrap generic CPU & RAM usage in a specific Flux struct.

jleeh (changes requested):

PR #27, opened 2021-07-14
Cronjob

skudasov:
Cronjob smoke test, creates cron job that runs every second and wait for 5 runs completed without errors

### DIFF
--- a/client/chainlink.go
+++ b/client/chainlink.go
@@ -31,6 +31,8 @@ type Chainlink interface {
 	ReadBridge(name string) (*BridgeType, error)
 	DeleteBridge(name string) error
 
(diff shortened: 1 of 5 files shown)


client/chainlink.go
```diff
@@ -31,6 +31,8 @@ type Chainlink interface {
 	ReadBridge(name string) (*BridgeType, error)
 	DeleteBridge(name string) error
 
+	ReadRunsForJob(jobID string) (*JobRunsResponse, error)
```
jleeh, line 34: Call this `ReadRunsByJob`

jleeh (changes requested):

jleeh (approved):

PR #30, opened 2021-07-21
Added basic direct request smoke test

skudasov:


client/chainlink_models.go
```diff
@@ -402,3 +405,23 @@ func ObservationSourceSpec(url string) string {
 			parse    [type=jsonparse path="data,result"];    
 			fetch -> parse;`, url)
 }
+
+// ObservationSourceConsumerTxEncode spec to put back "directrequest" job data on-chain
+func ObservationSourceConsumerTxEncode(url string) string {
```
jleeh, line 410: From looking at this, looks like it should be its own gotemplate object like the others.

jleeh (changes requested):

config/config.yml
```diff
@@ -5,7 +5,7 @@ networks:
     ethereum_hardhat:
         name: "Ethereum Hardhat"
         url: "ws://localhost:8545"
-        chain_id: 31337
+        chain_id: 1337
```
skudasov, line 8: Compatibility with geth/hh compose changes.

jleeh (approved):

PR #35, opened 2021-07-28
Feature/jh deployment ah version

kalverra:


environment/environment.go
```diff
@@ -456,13 +466,32 @@ func (env *k8sEnvironment) doPortForward(pod *coreV1.Pod, forwarder *portforward
 	}
 }
 
+func (env *k8sEnvironment) deployChainlinkSecrets() error {
```
kalverra, line 469: Kludge, tired of dealing with the weird secrets bug. Coming back to this in the A.M.

We might want something similar to this though. It's a bit more efficient to have all the nodes just use the same secret, less to deal with in handling secret names (using `generateName` makes us need to retrieve the final one after deployment), which might be a headache. Plus it's more efficient. 

The only issue would be that, well, we need to use the same secrets values for all nodes. Not sure how much that matters though, especially at this point.

environment/environment.go
```diff
@@ -511,8 +544,21 @@ func (k8s *K8sManifest) parse(path string, obj interface{}, previousManifests ma
 	if err != nil {
 		return fmt.Errorf("failed to read k8s template file %s: %v", path, err)
 	}
+
+	chainlinkConnectDetails := struct {
```
kalverra, line 548: I suspect you won't like this, but other attempts at getting a clean marshall in just `previousManifests` were super messy. Kicked around your idea for utilizing callback functions a bit, but didn't get far.

kalverra:
Environment test passing

PR #39, opened 2021-07-29
Remove the path specifications

kalverra:
Uses the `git` command line to get the project's root directory. This gets rid of specifying the path for launching chainlink envs.

```go
NewChainlinkCluster("../../", 1) // This is no more
NewChainlinkCluster(1) // New usage
```

### DIFF
--- a/internal/project_path.go
+++ b/internal/project_path.go
@@ -1,0 +1,12 @@
+package internal
+
+import (
+	"path/filepath"
+	"runtime"
+)
+
+var (
+	_, b, _, _ = runtime.Caller(0)
+	// Root folder of this project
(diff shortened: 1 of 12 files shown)


jleeh (approved):

jleeh (changes requested):
See Sergey's comment wrt Golang way, also can you add the same change for the config file?

internal/project_path.go
```diff
@@ -0,0 +1,12 @@
+package internal
```
jleeh, line 1: I'd put this in the `tools` package that's pre-existing.

jleeh (changes requested):

jleeh (approved):

PR #40, opened 2021-07-30
Fix tests on k8s

skudasov:
Fixed all the tests, refactored, direct request (runlog) test requires geth that will be added in a separate PR.

### DIFF
--- a/suite/contracts/contracts_runlog_test.go
+++ b/suite/contracts/contracts_runlog_test.go
@@ -3,12 +3,9 @@ package contracts
 import (
 	"context"
 	"errors"
-	"math/big"
-	"strings"
-
 	"github.com/avast/retry-go"
+	"github.com/ethereum/go-ethereum/common"
 	. "github.com/onsi/ginkgo"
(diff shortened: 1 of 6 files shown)


suite/contracts/contracts_runlog_test.go
```diff
@@ -18,87 +18,102 @@ import (
...
+				big.NewInt(1e18),
+				adapter.ClusterURL()+"/five",
+				"data,result",
+				big.NewInt(100),
+			)
+			Expect(err).ShouldNot(HaveOccurred())
+		})
+	})
+
+	Describe("Check directrequest job", func() {
```
jleeh, line 98: Small comment, but the description of steps with describe should not be directly stating "Check" or "Verify" etc. It should be like:
```
with DirectRequest job
```
then:
```
receives API call data on-chain
```

One weird thing I noticed in the Ginkgo docs is that the `By` calls all capitalised the first letter, whereas `Describe` and `It` didn't.

jleeh (changes requested):

jleeh (approved):

PR #46, opened 2021-08-02
Added geth deployment for k8s

skudasov:
Added geth dev network deployment for k8s, tried to run direct request test, works fine.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -4,41 +4,54 @@
 # KEEP_ENVIRONMENTS=OnFail
 # APPS_CHAINLINK_VERSION=1.0.0
 
-network: "ethereum_hardhat" # Selected network for test execution
+network: "ethereum_geth" # Selected network for test execution
 keep_environments: Never # Options: Always, OnFail, Never
 
 apps:
     chainlink:
+        image: smartcontract/chainlink
         version: 0.10.10
 
+private_keys: &private_keys
+    private_keys:
+        - ac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80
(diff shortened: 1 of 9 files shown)


environment/templates/geth-deployment.yml
```diff
@@ -0,0 +1,154 @@
...
+      containers:
+      - name: geth-network
+        image: ethereum/client-go:v1.10.3
+        entrypoint: sh ./root/init.sh
+        volumemounts:
+        - name: geth-configmap-volume
+          mountpath: /root/config
+        - name: geth-configmap-volume
+          mountpath: /root/.ethereum/devchain/keystore/key1
+          subpath: key1
```
skudasov, line 21: Trying to figure out why mounting the whole directory with `readonly: false` doesn't work, geth still has wrong access to that dir, tried 0777/0644.

config.yml
```diff
@@ -4,41 +4,54 @@
...
 
-network: "ethereum_hardhat" # Selected network for test execution
+network: "ethereum_geth" # Selected network for test execution
 keep_environments: Never # Options: Always, OnFail, Never
 
 apps:
     chainlink:
-        version: 0.10.10
+        image: public.ecr.aws/chainlink/chainlink
+        version: develop.a87e3d12686e0c157aa85cc557f66f0ebca0df25
```
jleeh, line 13: I wouldn't lock root config to a develop commit version, but I'm presuming you did this to avoid tests failing due to bugs in 0.10.10? I'd always keep locked to a release version, then just use env vars to run locally with the develop version.

jleeh (changes requested):
One small comment, but nice work on this PR.

jleeh (approved):

PR #50, opened 2021-08-03
Feature/OCR test fixes

kalverra:
# OCR Test Fixes

Fixes for OCR testing to work properly

## Building off Jonny's `ServiceDetails` Work

Simplifying what Jonny did for the `ServiceDetails` stuff so that P2P works. Didn't need quite all of the info all the time though. Just needed the remoteIP for the chainlink nodes to hang on to. This allows you to just pass in an array of `ChainlinkNode`s for configuration purposes, looks much cleaner.

```go
// GetChainlinkClients will return all instantiated Chainlink clients for a given environment
func GetChainlinkClients(env Environment) ([]client.Chainlink, error) {
	var clients []client.Chainlink

	sd, err := env.GetAllServiceDetails(ChainlinkWebPort)
	if err != nil {
		return nil, err
	}
	for _, service := range sd {
		linkClient, err := client.NewChainlink(&client.ChainlinkConfig{
			URL:      service.LocalURL.String(),
			Email:    "notreal@fakeemail.ch",
			Password: "twochains",
			RemoteIP: service.RemoteURL.Hostname(),
		}, http.DefaultClient)
		if err != nil {
			return nil, err
		}
		clients = append(clients, linkClient)
	}
	return clients, nil
}
```

## Changing How We Get Our Funding

```go
nativeAmount, linkAmount *big.Float
```

Funding was full of integer overflows due to confusion around how much we were sending. I've put in a new proposal where token values are listed, e.g.
```go
var OneLINK = big.NewFloat(1000000000000000000)
...
var OneEth = big.NewFloat(1000000000000000000)
```

And when calling to fund a contract or node, you simply give a multiple of one of that currency. e.g.
```go
err := actions.FundChainlinkNodes(
  chainlinkNodes,
  suiteSetup.Client,
  defaultWallet,
  big.NewFloat(2),
  big.NewFloat(2), // Fund each node with 2 ETH and 2 LINK each
)
...
err = fluxInstance.Fund(defaultWallet, nil, big.NewFloat(.005)) // Fund flux monitor contract with .005 ETH
```

### DIFF
--- a/client/chainlink.go
+++ b/client/chainlink.go
@@ -44,6 +44,7 @@ type Chainlink interface {
 	ReadETHKeys() (*ETHKeys, error)
 	PrimaryEthAddress() (string, error)
 
+	RemoteIP() string
 	SetSessionCookie() error
 
 	// Used for testing
@@ -238,6 +239,11 @@ func (c *chainlink) PrimaryEthAddress() (string, error) {
 		return "", err
 	}
 	return ethKeys.Data[0].Attributes.Address, nil
+}
+
+// RemoteIP retrieves the inter-cluster IP of the chainlink node, for use with inter-node communications
+func (c *chainlink) RemoteIP() string {
+	return c.Config.RemoteIP
 }
 
 // SetSessionCookie authenticates against the Chainlink node and stores the cookie in client state
--- a/client/chainlink.go
+++ b/client/chainlink.go
@@ -16,7 +16,7 @@ import (
 var ErrNotFound = errors.New("unexpected response code, got 404")
 var ErrUnprocessableEntity = errors.New("unexpected response code, got 422")
 
-var OneLINK = big.NewFloat(1000000000000000000)
+var OneLINK = big.NewFloat(10 ^ 18)
 
 // Chainlink interface that enables interactions with a chainlink node
 type Chainlink interface {
--- a/client/chainlink.go
+++ b/client/chainlink.go
@@ -86,7 +86,7 @@ func (c *chainlink) CreateJob(spec JobSpec) (*Job, error) {
 	if err != nil {
 		return nil, err
 	}
-	log.Info().Str("Node URL", c.Config.URL).Str("Job Body", specString).Msg("Creating Job")
+	log.Info().Str("Node URL", c.Config.URL).Msg("Creating Job")
 	_, err = c.do(http.MethodPost, "/v2/jobs", &JobForm{
 		TOML: specString,
 	}, &job, http.StatusOK)
--- a/client/chainlink.go
+++ b/client/chainlink.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
+	"math/big"
 	"net/http"
 	"strings"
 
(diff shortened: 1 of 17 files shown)


client/chainlink.go
```diff
@@ -15,6 +16,8 @@ import (
 var ErrNotFound = errors.New("unexpected response code, got 404")
 var ErrUnprocessableEntity = errors.New("unexpected response code, got 422")
 
+var OneLINK = big.NewFloat(1000000000000000000)
```
jleeh, line 19: 1^18 to make this cleaner, ^18 is always 1 Ether

client/ethereum.go
```diff
@@ -17,6 +17,8 @@ import (
 	"github.com/rs/zerolog/log"
 )
 
+var OneEth = big.NewFloat(1000000000000000000)
```
jleeh, line 20: Same here

environment/environment.go
```diff
@@ -19,30 +20,31 @@ type Environment interface {
...
 	if err != nil {
-		return nil, nil, err
+		return nil, err
 	}
 	for _, service := range sd {
 		linkClient, err := client.NewChainlink(&client.ChainlinkConfig{
 			URL:      service.LocalURL.String(),
 			Email:    "notreal@fakeemail.ch",
 			Password: "twochains",
+			RemoteIP: service.RemoteURL.Hostname(),
```
jleeh, line 40: This is where I'd say something like:
> The clients need to be unopinionated to how we deploy environments so it needs to be handled separately.

But it makes the tests much more cleaner, so makes sense.

client/chainlink_models.go
```diff
@@ -389,9 +390,55 @@ type OCRTaskJobSpec struct {
...
 
+type P2PData struct {
+	RemoteIP string
+	PeerID   string
+}
+
 func (o *OCRTaskJobSpec) Type() string { return "offchainreporting" }
 func (o *OCRTaskJobSpec) String() (string, error) {
-	fluxMonitorTemplateString := `type = "offchainreporting"
+	// Pre-process P2P data for easier templating
```
kalverra, line 400: This feels dirty, but doing it withing the template felt dirtier, open to suggestions on this

jleeh (changes requested):

client/chainlink_models.go
jleeh: I like it, putting a list of clients within the model for it then to be parsed in the templates avoids duplicating this for loop throughout a bunch of tests.

suite/refill/eth_refill_test.go
```diff
@@ -94,7 +94,7 @@ var _ = Describe("FluxAggregator ETH Refill", func() {
 		})
 
 		By("Funding ETH for a single round", func() {
-			err = actions.FundChainlinkNodes(nodes, s.Client, s.Wallets.Default(), big.NewInt(1e16), nil)
+			err = actions.FundChainlinkNodes(nodes, s.Client, s.Wallets.Default(), big.NewFloat(1), nil)
```
skudasov, line 97: Actually we need 1e16 here, just for one round.

skudasov (changes requested):

jleeh (approved):

skudasov (approved):

PR #53, opened 2021-08-05
Grabs logs from environment pods on test failure

kalverra:
# Grabbing Logs on Test Failure
On a test failure, creates a log folder for that test, then creates files for each pod and dumps logs into them.
logs_image

### DIFF
--- a/actions/setup.go
+++ b/actions/setup.go
@@ -1,6 +1,8 @@
 package actions
 
 import (
+	"os"
+	"path/filepath"
 	"strings"
 	"time"
 
@@ -12,6 +14,7 @@ import (
 	"github.com/smartcontractkit/integrations-framework/config"
 	"github.com/smartcontractkit/integrations-framework/contracts"
 	"github.com/smartcontractkit/integrations-framework/environment"
+	"github.com/smartcontractkit/integrations-framework/tools"
 )
 
 const (
@@ -79,6 +82,25 @@ func DefaultLocalSetup(
 }
 
 func (s *DefaultSuiteSetup) TearDown() func() {
+	if ginkgo.CurrentGinkgoTestDescription().Failed { // If a test fails, dump logs
+		logsFolder := tools.ProjectRoot + "/logs/"
+		if _, err := os.Stat(logsFolder); os.IsNotExist(err) {
+			if err = os.Mkdir(logsFolder, 0755); err != nil {
+				log.Err(err).Str("Log Folder", logsFolder).Msg("Error creating logs directory")
+			}
+		}
+		testLogFolder := filepath.Join(logsFolder, strings.Replace(ginkgo.CurrentGinkgoTestDescription().TestText, " ", "-", -1)+
+			"_"+s.Env.ID()+"/")
+		// Create specific test folder
+		if _, err := os.Stat(testLogFolder); os.IsNotExist(err) {
+			if err = os.Mkdir(testLogFolder, 0755); err != nil {
+				log.Err(err).Str("Log Folder", testLogFolder).Msg("Error creating logs directory")
+			}
+		}
+
+		s.Env.WriteLogs(testLogFolder)
+		log.Info().Str("Log Folder", testLogFolder).Msg("Wrote environment logs")
+	}
 	return func() {
 		switch strings.ToLower(s.Config.KeepEnvironments) {
 		case KeepEnvironmentsNever:
--- a/actions/setup.go
+++ b/actions/setup.go
@@ -1,10 +1,11 @@
 package actions
 
 import (
-	"github.com/onsi/ginkgo"
-	"github.com/rs/zerolog/log"
 	"strings"
 	"time"
+
+	"github.com/onsi/ginkgo"
+	"github.com/rs/zerolog/log"
 
 	"github.com/avast/retry-go"
 	"github.com/smartcontractkit/integrations-framework/client"
--- a/environment/k8s_environment.go
+++ b/environment/k8s_environment.go
@@ -211,12 +211,11 @@ func (env *k8sEnvironment) logFailures() {
 		if err != nil {
 			log.Err(err).Str("File Name", logFile.Name()).Msg("Error creating log file")
 		}
-		defer logFile.Close()
 
-		logFile.WriteString("======================================================================================\n")
-		logFile.WriteString("Test file name: " + ginkgo.CurrentGinkgoTestDescription().FileName + "\n")
-		logFile.WriteString("Test file line: " + fmt.Sprint(ginkgo.CurrentGinkgoTestDescription().LineNumber) + "\n")
-		logFile.WriteString("======================================================================================\n")
+		_, _ = logFile.WriteString("======================================================================================\n")
+		_, _ = logFile.WriteString("Test file name: " + ginkgo.CurrentGinkgoTestDescription().FileName + "\n")
+		_, _ = logFile.WriteString("Test file line: " + fmt.Sprint(ginkgo.CurrentGinkgoTestDescription().LineNumber) + "\n")
+		_, _ = logFile.WriteString("======================================================================================\n")
 
 		podLogOptions := &coreV1.PodLogOptions{}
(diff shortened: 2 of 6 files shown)


environment/k8s_environment.go
```diff
@@ -169,6 +177,69 @@ func (env *k8sEnvironment) TearDown() {
...
+			log.Err(err).Str("File Name", logFile.Name()).Msg("Error creating log file")
+		}
+
+		_, _ = logFile.WriteString("======================================================================================\n")
+		_, _ = logFile.WriteString("Test file name: " + ginkgo.CurrentGinkgoTestDescription().FileName + "\n")
+		_, _ = logFile.WriteString("Test file line: " + fmt.Sprint(ginkgo.CurrentGinkgoTestDescription().LineNumber) + "\n")
+		_, _ = logFile.WriteString("======================================================================================\n")
+
+		podLogOptions := &coreV1.PodLogOptions{}
+		if strings.Contains(podName, "chainlink") {
```
jleeh, line 221: Loop through all of the containers in a pod and then get all of their logs.

environment/k8s_environment.go
```diff
@@ -169,6 +177,69 @@ func (env *k8sEnvironment) TearDown() {
...
+	// Get logs from K8s pods
+	podsClient := env.k8sClient.CoreV1().Pods(env.namespace.Name)
+	podsList, err := podsClient.List(context.Background(), metaV1.ListOptions{})
+	if err != nil {
+		log.Err(err).Str("Env Name", env.namespace.Name).Msg("Error retrieving pod list from K8s environment")
+	}
+
+	// Each Pod gets a file
+	for _, pod := range podsList.Items {
+		podName := pod.Labels[SelectorLabelKey]
```
jleeh, line 209: Just to shorten this function, take this out of here and have another internal function that is like:
```
func getLogsForPod(pod) ([]byte, error)
```

environment/k8s_environment.go
```diff
@@ -158,6 +163,9 @@ func (env *k8sEnvironment) GetServiceDetails(remotePort uint16) (*ServiceDetails
 // TearDown cycles through all the specifications and tears down the deployments. This typically entails cleaning
 // up port forwarding requests and deleting the namespace that then destroys all definitions.
 func (env *k8sEnvironment) TearDown() {
+	if ginkgo.CurrentGinkgoTestDescription().Failed { // If a test fails, dump logs
```
jleeh, line 166: To avoid needing to import Ginkgo in environments, split out this function. Have a new `ExportLogsToFolder(folder string, logPrefix log.Buffer)` function in the environment interface, that way Ginkgo isn't needed and the logic to whether logs are needed to be written is still handled in the `actions` package.

To replicate the functionality where you output a heading to the log file with more info, you can then instantiate the buffer in teardown with that output and simply copy the buffer each time and append to it.

environment/k8s_environment.go
```diff
@@ -169,6 +177,69 @@ func (env *k8sEnvironment) TearDown() {
 	log.Error().Err(err)
 }
 
+// On a test failure, dump all the pod logs
+func (env *k8sEnvironment) logFailures() {
+	// Create logs directory if non-existent
+	log.Info().Msg("Creating log files on test failure")
+	logsFolder := tools.ProjectRoot + "/logs/"
```
jleeh, line 184: I'd probably also put the creation of the logs folder within TearDown in actions as this would be a common function amongst all implementations of environments.

environment/k8s_environment.go
```diff
@@ -169,6 +177,69 @@ func (env *k8sEnvironment) TearDown() {
...
+		buf := new(bytes.Buffer)
+		_, err = io.Copy(buf, podLogs)
+		if err != nil {
+			log.Err(err).Str("Env Name", env.namespace.Name).Msg("error in copying information from podLogs to a buffer")
+		}
+		_, _ = logFile.Write(buf.Bytes())
+
+		_ = logFile.Close()
+		_ = podLogs.Close()
+	}
```
jleeh, line 240: Have a log line that states `Logs written to file` with the namespace and file path.

environment/k8s_environment.go
```diff
@@ -169,6 +177,69 @@ func (env *k8sEnvironment) TearDown() {
...
+		podLogRequest := podsClient.GetLogs(pod.Name, podLogOptions)
+		podLogs, err := podLogRequest.Stream(context.Background())
+		if err != nil {
+			log.Err(err).Str("Env Name", env.namespace.Name).Msg("error in opening pod logging stream")
+		}
+
+		buf := new(bytes.Buffer)
+		_, err = io.Copy(buf, podLogs)
+		if err != nil {
+			log.Err(err).Str("Env Name", env.namespace.Name).Msg("error in copying information from podLogs to a buffer")
```
jleeh, line 234: k8s specific, but logging has been consistent in using `Namespace` for the key in the log output.

jleeh (changes requested):

kalverra:
The new style gets all logs from all containers, thus creates a folder for each pod. Logic to condense it for cases where there's only 1 container was a bit lengthy, will gauge the temp on this solution.
logs_image

environment/k8s_environment.go
kalverra: Followed this, but went a little simpler with it, will see how you like it.

environment/k8s_environment.go
```diff
@@ -169,6 +199,38 @@ func (env *k8sEnvironment) TearDown() {
...
+// Writes logs for each container in a pod
+func writeLogsForPod(podsClient v1.PodInterface, pod coreV1.Pod, podFolder string) error {
+	for _, container := range pod.Spec.Containers {
+		logFile, err := os.Create(filepath.Join(podFolder, container.Name) + ".log")
+		if err != nil {
+			log.Err(err).Str("File Name", logFile.Name()).Msg("Error creating log file")
+		}
+
+		_, _ = logFile.WriteString("======================================================================================\n")
+		_, _ = logFile.WriteString("Test file name: " + ginkgo.CurrentGinkgoTestDescription().FileName + "\n")
```
jleeh, line 211: The point of the log buffer was to remove the need for importing Ginkgo in the environment, as this package should not be tied to the implementation of tests.

jleeh (changes requested):

suite/refill/eth_refill_test.go
```diff
@@ -16,6 +16,37 @@ import (
 	"github.com/smartcontractkit/integrations-framework/environment"
 )
 
+var _ = FDescribe("TODO", func() {
```
skudasov, line 19: Don't forget to move that to another place when all is done =)

environment/k8s_environment.go
kalverra: Forgot to rip this out, already have this info in our unit reports

actions/setup.go
```diff
@@ -78,6 +82,25 @@ func DefaultLocalSetup(
 }
 
 func (s *DefaultSuiteSetup) TearDown() func() {
+	if ginkgo.CurrentGinkgoTestDescription().Failed { // If a test fails, dump logs
+		logsFolder := tools.ProjectRoot + "/logs/"
+		if _, err := os.Stat(logsFolder); os.IsNotExist(err) {
```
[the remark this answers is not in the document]
kalverra: Too many import cycles, but I feel you. Might do a `helpers` package refactor?

environment/k8s_environment.go
```diff
@@ -169,6 +198,33 @@ func (env *k8sEnvironment) TearDown() {
...
+		if err != nil {
+			return err
+		}
+
+		buf := new(bytes.Buffer)
+		_, err = io.Copy(buf, podLogs)
+		if err != nil {
+			return err
+		}
+		_, _ = logFile.Write(buf.Bytes())
```
jleeh, line 220: Since the func returns an error, you should error handle here and below.

jleeh (changes requested):

jleeh (approved):

PR #54, opened 2021-08-06
Support for parallel transactions

jleeh:
Revamped transaction management in the Ethereum client, also adding new functions to the Blockchain client interface for toggling parallelising transactions, clean shutdowns and a blocking function for waiting for transactions.

By default, parallelisation is disabled. It can be turned on by calling:
```
blockchainClient.ParallelTransactions(true)
```

The Ethereum client will now also only initiate one subscription for block headers with subscription interfaces to receive any block headers, enabling tx confirmations.

This functionality is especially useful for tests with large environments, eg volume tests.

As part of this PR I've added support for Ganache, but it breaks with the Chainlink node with context timeout errors that I haven't fully debugged. PR works fine with Geth dev.

### DIFF
--- a/client/ethereum.go
+++ b/client/ethereum.go
@@ -4,8 +4,10 @@ import (
 	"context"
 	"fmt"
 	"github.com/ethereum/go-ethereum/accounts/abi"
+	"github.com/hashicorp/go-multierror"
 	"github.com/pkg/errors"
 	"math/big"
+	"sync"
 	"time"
 
 	ethContracts "github.com/smartcontractkit/integrations-framework/contracts/ethereum"
@@ -25,6 +27,12 @@ var OneEth = big.NewFloat(1e18)
 type EthereumClient struct {
 	Client  *ethclient.Client
 	Network BlockchainNetwork
+
+	txQueue            chan common.Hash
+	queuedTransactions map[string]EthereumTransactionConfirmer
+	mutex              *sync.Mutex
+	queueTransactions  bool
+	doneChan           chan struct{}
 }
 
 // ContractDeployer acts as a go-between function for general contract deployment
@@ -42,16 +50,37 @@ func NewEthereumClient(network BlockchainNetwork) (*EthereumClient, error) {
 		return nil, err
 	}
 
-	return &EthereumClient{
-		Network: network,
-		Client:  cl,
-	}, nil
+	ec := &EthereumClient{
+		Network:            network,
+		Client:             cl,
+		txQueue:            make(chan common.Hash, 64), // Max buffer of 64 tx
+		queuedTransactions: map[string]EthereumTransactionConfirmer{},
+		mutex:              &sync.Mutex{},
+		queueTransactions:  false,
+		doneChan:           make(chan struct{}),
+	}
+	go ec.newHeadersLoop()
+	return ec, err
+}
+
+// Close tears down the current open Ethereum client
+func (e *EthereumClient) Close() error {
+	e.doneChan <- struct{}{}
+	e.Client.Close()
+	return nil
 }
 
 // Get returns the underlying client type to be used generically across the framework for switching
 // network types
 func (e *EthereumClient) Get() interface{} {
 	return e
+}
+
+// ParallelTransactions when enabled, sends the transaction without waiting for transaction confirmations. The hashes
+// are then stored within the client and confirmations can be waited on by calling WaitForTransactions.
+// When disabled, the minimum confirmations are waited on when the transaction is sent, so parallelisation is disabled.
+func (e *EthereumClient) ParallelTransactions(enabled bool) {
+	e.queueTransactions = enabled
 }
 
 // Fund funds a specified address with LINK token and or ETH from the given wallet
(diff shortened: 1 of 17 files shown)


client/ethereum.go
```diff
@@ -339,3 +406,88 @@ func (e *EthereumClient) errorReason(b ethereum.ContractCaller, tx *types.Transa
...
+		Str("Block Hash", block.Hash().Hex()).
+		Str("Block Number", block.Number().String()).Str("Tx Hash", t.txHash.Hex()).
+		Int("Minimum Confirmations", t.minConfirmations)
+	isConfirmed, err := t.eth.isTxConfirmed(t.txHash)
+	if err != nil {
+		return err
+	} else if isConfirmed {
+		t.confirmations++
+	}
+	if t.confirmations == t.minConfirmations {
```
kalverra, line 460: Worth an `>=` I think, just in case.

environment/templates/ganache-deployment.yml
```diff
@@ -0,0 +1,26 @@
...
+        app: ganache-network
+    spec:
+      containers:
+      - name: ganache-network
+        image: trufflesuite/ganache-cli:v6.12.2
+        ports:
+        - containerport: {{ .Values.evm.rpcPort }}
+        args:
+          - -d
+          - myth like bonus scare over problem client lizard pioneer submit female collect
```
kalverra, line 16: Some of these secret key phrases read like esoteric short stories or poems

client/ethereum.go
```diff
@@ -42,10 +50,24 @@ func NewEthereumClient(network BlockchainNetwork) (*EthereumClient, error) {
...
+	}
+	go ec.newHeadersLoop()
+	return ec, err
+}
+
+// Close tears down the current open Ethereum client
+func (e *EthereumClient) Close() error {
+	e.doneChan <- struct{}{}
+	e.Client.Close()
+	return nil
```
kalverra, line 70: Why only return nil? Shouldn't it return the err of `e.Client.Close()`?

kalverra (commented):
My overall concern with this style is that I am sure people will often forget to wait - especially after function calls that don't obviously yell "THIS IS A TRANSACTION" at them - and be locked in debug hell for a bit. I'm trying to think of ways to make this a bit more idiot-proof. 

Like each function that creates a transaction accepts a boolean `parallel` parameter, then returns a `wait` channel? You could collect them all and send them to 
```go
WaitForTransactions(confirmed chan bool...) error
```
or something like that. Though I feel that approach is likely a bit too verbose, even for go.

I will keep thinking on it, also will wait on Sergey to take a look as he's been digging around in this area more than I.

config.yml
```diff
@@ -54,6 +54,35 @@ networks:
         transaction_timeout: 1s
         minimum_confirmations: 1
         gas_estimation_buffer: 0
+    ethereum_ganache:
+        name: "Ethereum Ganache"
+        chain_id: 1337
+        type: evm
+        private_keys:
+            - 4f3edf983ac636a65a842ce7c78d9aa706d3b113bce9c46f30d7d21715b23b1d
```
skudasov, line 62: Can add keys from here and just reuse as anchor https://github.com/smartcontractkit/chainlink-node-compose/blob/main/docker-compose/services.yml

environment/templates/ganache-deployment.yml
skudasov: Can import already used keys

suite/contracts/contracts_flux_test.go
```diff
@@ -46,6 +48,8 @@ var _ = Describe("Flux monitor suite", func() {
 			Expect(err).ShouldNot(HaveOccurred())
 			err = fluxInstance.UpdateAvailableFunds(context.Background(), s.Wallets.Default())
 			Expect(err).ShouldNot(HaveOccurred())
+			err = s.Client.WaitForTransactions()
```
skudasov, line 51: Does it speed up the test? Do we need parallel here?

skudasov (changes requested):

client/ethereum.go
```diff
@@ -25,6 +27,12 @@ var OneEth = big.NewFloat(1e18)
 type EthereumClient struct {
 	Client  *ethclient.Client
 	Network BlockchainNetwork
+
```
skudasov, line 30: Would you mind to add nonce handling there too? Or I can merge it from volume branch as an additional option later

client/ethereum.go
jleeh: Did this deliberately as if it's `>=` it can show more confirmations than min confirmations until you call `Wait` which can be confusing. It's impossible for it to be greater than min without being equal.

client/ethereum.go
jleeh: `Close` has no returns, surprised me as well.

client/ethereum.go
jleeh: I'd merge it from the volume branch imo. I don't want to cherry-pick/reimplement and then have to deal with conflicts when you already have it working.

jleeh:
@kalverra You raise a good point and it's something I was thinking about when working on the feature. The initial implementation I had was `SendTransaction` had a bool parameter on which you can toggle whether to synchronously send the transaction or not. The problem with that was that within all the contract deployer functions you'd also need to cascade down the same parameter or all contract deployments would be async.

Managing channels in tests would up the tests complexity and would be prime for completely locking tests, having it all managed in the client with a timeout context is preferred as you know that if a tx doesn't confirm, or if there's no tx to confirm it'll never block. The `Wait` func imo is the lesser of the evils with this solution and is the simplest, especially for test scripting. If you don't want to concurrently send tx, then there's no change to existing tests.

client/ethereum.go
skudasov: Good, just asking if you ok that I'll add nonce handling to that code later.

skudasov (approved):

PR #58, opened 2021-08-09
Expand OCR test time

kalverra:
Quick fix for consistency issues. Talking to @jleeh about getting a blockchain method to subscribe to new blocks, to make this cleaner and more consistent than the basic check-sleep loop.

### DIFF
--- a/suite/contracts/contracts_ocr_test.go
+++ b/suite/contracts/contracts_ocr_test.go
@@ -14,7 +14,7 @@ import (
 	"github.com/smartcontractkit/integrations-framework/environment"
 )
 
-var _ = Describe("OCR Feed", func() {
+var _ = FDescribe("OCR Feed", func() {
 	var (
 		suiteSetup     *actions.DefaultSuiteSetup
 		chainlinkNodes []client.Chainlink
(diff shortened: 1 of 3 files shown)


tateexon (approved):
Looks good to me

suite/contracts/contracts_ocr_test.go
```diff
@@ -115,7 +115,7 @@ var _ = Describe("OCR Feed", func() {
 			Expect(err).ShouldNot(HaveOccurred())
 
 			// Wait for a round
-			for i := 0; i < 10; i++ {
+			for i := 0; i < 60; i++ {
```
jleeh, line 118: Use the retry lib that Sergey brought in, like here:
https://github.com/smartcontractkit/integrations-framework/blob/main/suite/contracts/contracts_runlog_test.go#L101

jleeh (changes requested):

kalverra:
# Improved OCR Test
Improved this to utilize the new block header subscription to poll only for new rounds, and included a value change check. The previous test time was ~60-70 seconds, this one includes another check, and consistently runs in ~70-75 seconds

suite/contracts/contracts_ocr_test.go
```diff
@@ -101,42 +101,47 @@ var _ = Describe("OCR Feed", func() {
 					P2PBootstrapPeers:  []client.Chainlink{bootstrapNode},
 					KeyBundleID:        nodeOCRKeyId,
 					TransmitterAddress: nodeTransmitterAddress,
-					ObservationSource:  client.ObservationSourceSpec(adapter.ClusterURL() + "/five"),
+					ObservationSource:  client.ObservationSourceSpec(fmt.Sprintf("%s/variable", adapter.ClusterURL())),
```
jleeh, line 104: I'd add this endpoint within the adapter interface so we don't need to do this string formatting each time.

jleeh (changes requested):

jleeh (approved):

PR #62, opened 2021-08-11
Added Explorer mock checks

skudasov:
Explorer OCR telemetry mock service. Refactored HTTP client for reuse.


environment/templates/explorer-deployment.yml
```diff
@@ -0,0 +1,23 @@
...
+spec:
+  selector:
+    matchlabels:
+      app: explorer-mock
+  template:
+    objectmeta:
+      labels:
+        app: explorer-mock
+    spec:
+      containers:
```
tateexon, line 12: Do we want to add request/limit values into this spec for the cpu and memory so we avoid adding another unbound deployment?

skudasov: Nice catch, will figure out some limits for it.

skudasov: Currently, we have issues with proper YAML unmarshalling of manifests, think we need to settle them down first.

skudasov: Done

jleeh (approved):

PR #64, opened 2021-08-11
Prepare Config for Use in Chainlink Repo

kalverra:
# Config Loading Options

We had a similar style to this config loading setup earlier on in the project; we moved away from it because it was annoying having to specify the config. Well, we're back. The idea is to move the test suite into the main chainlink repo, and in order for things like reading the config, and recording log files properly, there doesn't seem to be many ways around just providing the config path. 

## Other Attempts

I open the floor for discussion, as I've tried a few different ways to do this, and they didn't pan out. All attempts to automatically sense the directory from where the calls were being made led to one of the following.

1. It kept defaulting to the `integrations-framework` library
2. The root directory was assumed to be some compiled Ginkgo binary in a randomly generated `var` folder
3. The calling file was an internal `Ginkgo` runner
4. Getting the current working directory, which is obviously great if you are calling tests from the proper directory. Not such a great solution otherwise.

So I'm left at this option, which works fine, tests are running in the chainlink repo. But it feels icky.

## Usage

Addition of this setup action
```go
// DefaultLocalSetupSpecifyConfig gives a default testing environment setup with a specified config file path
func DefaultLocalSetupSpecifyConfig(
	envInitFunc environment.K8sEnvSpecInit,
	initFunc client.BlockchainNetworkInit,
	configPath string,
) (*DefaultSuiteSetup, error) {
	s, err := DefaultLocalSetup(envInitFunc, initFunc)
	if err != nil {
		return nil, err
	}
	specifiedConf, err := config.NewConfigWithPath(config.LocalConfig, configPath)
	if err != nil {
		return nil, err
	}
	s.Config = specifiedConf
	return s, err
}
```

Example usage in a `BeforeEach` block I've setup for the main chainlink repo:

```go
BeforeEach(func() {
  confFileLocation, err = filepath.Abs("../") // Get the absolute path of the test suite's root directory
  suiteSetup, err = actions.DefaultLocalSetupSpecifyConfig(
	  environment.NewChainlinkCluster(1),
	  client.NewNetworkFromConfig,
	  confFileLocation,
  )
})
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,50 @@ The framework is primarilly intended to facillitate testing chainlink features a
 This framework is still very much a work in progress, and will have frequent changes, many of which will probably be
 breaking.
 
+## Usage
+
+```go
+var _ = Describe("Basic Contract Interactions", func() {
+	var suiteSetup *actions.DefaultSuiteSetup
+	var defaultWallet client.BlockchainWallet
+
+	BeforeEach(func() {
+		By("Deploying the environment", func() {
+			confFileLocation, err := filepath.Abs("../") // Get the absolute path of the test suite's root directory
+            Expect(err).ShouldNot(HaveOccurred())
+			suiteSetup, err = actions.DefaultLocalSetup(
+				environment.NewChainlinkCluster(0),
+				client.NewNetworkFromConfig,
+				tools.ProjectRoot, // Folder location of the
+			)
+			Expect(err).ShouldNot(HaveOccurred())
+			defaultWallet = suiteSetup.Wallets.Default()
+		})
+	})
+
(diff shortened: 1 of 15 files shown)


tateexon:
The example usage and code looks good. I think we should probably move your example usage into the readme.

Should we be adding unit tests to our test framework pieces like the config?

kalverra:
@tateexon Thanks, I agree on adding examples to the README, will put that in with this branch. Also yes on the unit tests, we've been holding off on large amounts of unit testing while the project is still in so much flux, but we're probably hitting a stable enough point now to start adding them.

tateexon (approved):

PR #69, opened 2021-08-17
Add go bindings build for contracts

skudasov:
Contract bindings generator for different sources, an example config:
```
contracts:
    ethereum:
        executable_path: ./tools/bin/abigen
        out_path: ./contracts/ethereum
        sources:
            local:
                path: ../chainlink/contracts/artifacts/src
            external:
                path: ./external_artifacts
                region: us-west-2
                s3_path: contract-artifacts-test-upload
                repositories:
                    keeper:
                        path: keeper/artifacts/contracts
                        commit: 19ba7cfbe7b7c6ca7621b090e229b4ba1147e2e1
                    ocr:
                        path: ocr/artifacts/contract/src
                        commit: f27c14a905c5735abbb6e0c9699e9d0e3e9b7217
```
The whole goal of that refactoring is that IMO we need only to generate bindings, not to build contracts, so we can add Hardhat build steps to all our repositories, and every team can decide how and when they release contracts, and we can pull their versions to check integration. 

We split all our contracts into 3 categories for every network(language):
1. Contracts that we wrote are only useful in integration tests, we can add them to Chainlink main repo under `integration` folder, just not to duplicate Hardhat build configuration in the main repo.
2. Local contracts. After we migrate to the main Chainlink repo we can use their Hardhat build without modifications, run `hardhat compile` in CI in branch to build them, generate go bindings for our tests, and run the integration suite.
3. External contracts. For example https://github.com/smartcontractkit/offchain-reporting and https://github.com/smartcontractkit/keeper we can add Hardhat compilation step and upload their artifacts to S3 in CI, after that test integration against a particular commit, the tool will download and unzip Hardhat artifacts to generate bindings on build step.

Usage:
```
make install_cli
ifcli build_contracts -c config.yml
```


README.md
```diff
@@ -96,3 +102,15 @@ Run:
 ```
 go test ./..
 ```
+
+### Build contracts
+Example of generating go bindings for Ethereum contracts
+```
+ifcli build_contracts -e ${ABIGEN_BINARY}
```
skudasov, line 109: Chainlink's main repo already has abigen in `tools/bin/abigen`, maybe we can reuse it after merging repos?

contracts/build.go
```diff
@@ -0,0 +1,127 @@
...
+
+// NewEthereumContractBuilder builds solidity contracts
+func NewEthereumContractBuilder(execPath string) *EthereumContractBuilder {
+	return &EthereumContractBuilder{ExecutablePath: execPath}
+}
+
+// Compile compiles contracts
+func (b *EthereumContractBuilder) Compile(_ []*ContractTarget) error {
+	// step is not needed for Ethereum, already has ABI/BIN build in CI
+	return nil
```
skudasov, line 51: We already have contracts compilation there https://github.com/smartcontractkit/chainlink/blob/develop/.github/workflows/continuous-integration-workflow.yml#L67 can skip this step and generate bindings from ABI/BIN only.

contracts/build.go
```diff
@@ -0,0 +1,127 @@
...
+	newContents := strings.Replace(string(read), pkg, DefaultBindPkg, 1)
+	err = ioutil.WriteFile(filepath, []byte(newContents), 0)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// abigen calls abigen binary to generate bindings
+func abigen(a abigenArgs) error {
```
skudasov, line 109: Can use abigen programmatically, but need to keep in mind go-ethereum version in main repo.

tools/project_path.go
```diff
@@ -6,6 +6,9 @@ import (
 )
 
 var (
-	_, b, _, _  = runtime.Caller(0)
-	ProjectRoot = filepath.Join(filepath.Dir(b), "/..")
+	_, b, _, _ = runtime.Caller(0)
+	// ProjectRoot Root folder of this project
+	ProjectRoot          = filepath.Join(filepath.Dir(b), "/..")
+	ContractsDir         = filepath.Join(ProjectRoot, "contracts")
```
skudasov, line 12: Those paths can be changed after repo migration, or added as a CLI param.

jleeh (approved):

PR #72, opened 2021-08-19
FM deviation flag test

skudasov:
Flags/DeviationFlaggingValidator/ReadAccessController/WriteAccessController contracts and bindings.
Add external validator test for FM, we already have such a test in the main repo, but it'll be nice not to forget to test it with other networks.


suite/contracts/contracts_flux_validate_flag_test.go
```diff
@@ -0,0 +1,155 @@
...
+				}
+				_, err = n.CreateJob(fluxSpec)
+				Expect(err).ShouldNot(HaveOccurred())
+			}
+		})
+	})
+
+	Describe("with Flux job", func() {
+		It("Sets a flag when value is above threshold", func() {
+			{
```
jleeh, line 127: I don't like this notation of using brackets in which I presume is organising the code. Just have it all on the same level.

jleeh (changes requested):

jleeh (approved):

PR #73, opened 2021-08-19
Feature 14051 Convert most tests to use bridge tasks instead of http tasks

tateexon:


client/chainlink_models.go
```diff
@@ -256,18 +261,16 @@ observationSource = """
...
+	DataPath             string
 }
 
 func (d *PipelineSpec) Type() string {
 	return "common_pipeline"
 }
 
 func (d *PipelineSpec) String() (string, error) {
-	sourceString := `fetch    [type=http method={{.Method}} url="{{.URL}}" requestData="{{.RequestData}}"];
+	sourceString := d.BridgeTypeAttributes.Task() + `
```
kalverra, line 273: Use `fmt.Sprintf` for concatenation as much as possible.

kalverra (changes requested):

suite/refill/eth_refill_test.go
```diff
@@ -81,15 +81,25 @@ var _ = Describe("FluxAggregator ETH Refill", func() {
...
+
+			os := &client.PipelineSpec{
+				BridgeTypeAttributes: bta,
+				DataPath:             "data,result",
 			}
 			ost, err := os.String()
 			Expect(err).ShouldNot(HaveOccurred())
+
 			for _, n := range nodes {
+				// create the brige
```
jleeh, line 99: Typo, but I always try to avoid comments for code that is pretty self explanatory. Comment is create the bridge, the function call below is `CreateBridge`.

suite/contracts/contracts_runlog_test.go
```diff
@@ -58,20 +60,31 @@ var _ = Describe("Direct request suite", func() {
...
-				Method:      "GET",
-				RequestData: "{}",
-				DataPath:    "data,result",
+				BridgeTypeAttributes: bta,
+				DataPath:             "data,result",
 			}
 			ost, err := os.String()
 			Expect(err).ShouldNot(HaveOccurred())
+
+			// create the job
```
jleeh, line 87: Same here

suite/refill/eth_refill_test.go
```diff
@@ -81,15 +81,25 @@ var _ = Describe("FluxAggregator ETH Refill", func() {
 		})
 
 		By("Adding FluxAggregator jobs to nodes", func() {
-			os := &client.PipelineSpec{
+			// create the bridge
```
jleeh, line 84: Here

suite/contracts/contracts_runlog_test.go
```diff
@@ -19,7 +19,7 @@ import (
 	"github.com/smartcontractkit/integrations-framework/tools"
 )
 
-var _ = Describe("Direct request suite", func() {
+var _ = Describe("Direct request suite @runlog", func() {
```
jleeh, line 22: 👍

client/chainlink_models.go
```diff
@@ -585,8 +586,16 @@ observationSource = """
 	return marshallTemplate(w, "Webhook Job", webHookTemplateString)
 }
 
-func ObservationSourceSpec(url string) string {
+// ObservationSourceSpecHttp creates a http GET task spec for json data
+func ObservationSourceSpecHttp(url string) string {
```
jleeh, line 590: I think this is a Go thing, but with acronyms you typically capitalise them all. So `HTTP`, `URL` etc.

jleeh (changes requested):

client/chainlink_models.go
skudasov: Can't we use templating here?

tateexon: I will look into using templating. Does templating support something like {{.BridgeTypeAttributes.RequestData}} where you need to go multiple levels down?

tateexon: Looks like templates do support that, should be an easy update.

jleeh (approved):

kalverra (approved):

PR #74, opened 2021-08-20
Feature/mixed versions

kalverra:
# Enable Running Mixed Chainlink Versions
This adds a new chainlink cluster deployment function
```go
environment.NewMixedVersionChainlinkCluster(5)
```

This is handy, as only a few of our tests need this sort of env setup, so we can call them case-by-case instead of wasting time running the suite once with mixed versions and then again with all the same. See the new OCR tests. 
```go
var _ = Describe("OCR Feed", func() {

	DescribeTable("Deploy and watch an OCR feed", func(
		suiteInit environment.K8sEnvSpecInit,
	) {
		var (
			...
		)

			...
		})

		By("Funding nodes and deploying OCR contract", func() {
			...
		})

		By("Sending OCR jobs to chainlink nodes", func() {
			...
		})

		By("Checking OCR rounds", func() {
			...
		})

		By("Checking explorer telemetry", func() {
			...
		})

		By("Tearing down the environment", suiteSetup.TearDown())
	},
		Entry("all the same version", environment.NewChainlinkCluster(5)),
		Entry("different versions", environment.NewMixedVersionChainlinkCluster(5)),
	)
})
```

## How It's Done
We manipulate the `values["chainlink"]` store to indicate that we want different versions. When you ask for a multi-version deployment, we query the latest 2 releases from GitHub and mix them in with whatever the default config release is. We then have our deployment template check on the existence of these values, then run some template functions to handle it properly.
```
{{ if .Values.chainlinkCluster.chainlink.images }}{{ iterateImage .Values.chainlinkCluster.chainlink.images }}{{ else }}{{ .Config.Apps.Chainlink.Image }}{{ end }}:{{ if .Values.chainlinkCluster.chainlink.versions }}{{ iterateVersion .Values.chainlinkCluster.chainlink.versions }}{{ else }}{{ .Config.Apps.Chainlink.Version }}{{ end }}
```
It's a bit of a doozy.

## Issues
The whole `values` portion of environment deployment is a tad messy and leads you down assumptions that end up being wrong when trying to implement things like this. Now that I've got ahold of template functions, I think we could clean it up so that our templates pull `values` from the proper manifests, instead of just the `values["chainlink"]` map, which would make my brain smile; I'm dubious about how much reward past that there is for the effort made. Something to consider if we ever want to dig into this portion of the code again.


environment/environment_templates.go
```diff
@@ -173,6 +204,60 @@ func NewChainlinkCluster(nodeCount int) K8sEnvSpecInit {
 		manifests: manifests,
 	}
 
+	return newChainlinkManifest(chainlinkCluster, "basic-chainlink")
+}
+
+// NewMixedVersionChainlinkCluster mixes the currently latest chainlink version (as defined by the config file) with
+// the past two stable versions, ensuring that at least one of each is deployed
+func NewMixedVersionChainlinkCluster(nodeCount int) K8sEnvSpecInit {
```
tateexon, line 212: We should add unit tests for this function. It may also make sense to break the github client part out into its own block to make testing easier (mocking a function instead of mocking a client)

environment/templates/chainlink-deployment.yml
```diff
@@ -9,7 +9,7 @@ spec:
           secretName: {{ .Manifest.Secret.ObjectMeta.Name }}
       containers:
       - name: node
-        image: {{ .Config.Apps.Chainlink.Image }}:{{ .Config.Apps.Chainlink.Version }}
+        image: {{ if .Values.chainlinkCluster.chainlink.images }}{{ iterateImage .Values.chainlinkCluster.chainlink.images }}{{ else }}{{ .Config.Apps.Chainlink.Image }}{{ end }}:{{ if .Values.chainlinkCluster.chainlink.versions }}{{ iterateVersion .Values.chainlinkCluster.chainlink.versions }}{{ else }}{{ .Config.Apps.Chainlink.Version }}{{ end }}
```
tateexon, line 12: Logic being added to deployment files should be avoided if possible since it can be harder to follow the logic when you have to jump out of the code. Would it be easier to update the config values used before the deployment file is read and used?

tateexon (changes requested):

environment/k8s_environment.go
```diff
@@ -727,6 +738,14 @@ func (m *K8sManifest) setValues() error {
 	return nil
 }
 
+// Indexes to keep track of which versions have been deployed already
+var (
+	imageIndex   int = 0
+	versionIndex int = 0
+	imageMutex   sync.Mutex
```
tateexon, line 745: Are these mutex's used anywhere?

kalverra: Nope, good catch, removing

tateexon (approved):
Looks good to me

PR #77, opened 2021-08-23
Chore 14754 add unit tests for the config code

tateexon:


config/config_suite_test.go
```diff
@@ -0,0 +1,13 @@
+package config_test
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func TestConfig(t *testing.T) {
```
kalverra, line 10: Doesn't look like you use the logger stuff in these, but we'll want to keep the unit reporter. This is no longer necessary come Ginkgo 2.0, but until then...

```go
func TestContracts(t *testing.T) {
	RegisterFailHandler(Fail)
	log.Logger = log.Output(zerolog.ConsoleWriter{Out: os.Stderr})
	junitReporter := reporters.NewJUnitReporter("junit.xml")
	RunSpecsWithDefaultAndCustomReporters(t, "Environment Suite", []Reporter{junitReporter})
}
```

kalverra (changes requested):

config/config_suite_test.go
tateexon: Good catch! Fixed

kalverra (approved):

PR #89, opened 2021-09-08
Add reorg test

skudasov:
Blocks reorganization example test for Ethereum.
Running `runlog` deployment, forking the network, joining, and waiting until blocks starting from fork block get reorged.
Also added Loki dashboard to get some insights.
image
If there was no reorg on block interval [fork, latest] the test will fail.
```
NETWORK="ethereum_geth_reorg" ginkgo -r --focus="@reorg"
```


Makefile
```diff
@@ -8,7 +8,7 @@ lint: ## run linter
 
 .PHONY: golangci
 golangci: ## install golangci-linter
-	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b ${BIN_DIR} v1.41.1
+	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b ${BIN_DIR} v1.42.0
```
skudasov, line 11: Just to mitigate some discrepancies between local and CI builds.

client/ethereum.go
```diff
@@ -411,17 +428,32 @@ func (e *EthereumClient) subscribeToNewHeaders() error {
...
 		case header := <-headerChannel:
+			e.recordHeader(header)
 			e.receiveHeader(header)
 		case <-e.doneChan:
 			return nil
 		}
 	}
 }
 
+// recordHeader saves header either to first seen headers or to reorged/duplicated
```
skudasov, line 439: Can use that later to check if a network really receives reorged blocks.

contracts/ethereum_reorg_util.go
```diff
@@ -27,28 +27,36 @@ func AwaitMining(c client.BlockchainClient) error {
...
 }
 
-func (f *NextBlockConfirmer) ReceiveBlock(_ *types.Block) error {
+func (f *NextBlockConfirmer) ReceiveBlock(b *types.Block) error {
 	if f.done {
 		return nil
 	}
-	log.Info().Msg("First block received")
+	// in case of a multi-node setup we may need a block from the node on which reorg will be performed
+	if f.extraData != nil {
```
skudasov, line 54: Must wait for tx node to mine its own block, so the network is joined before splitting.

suite/chaos/reorg_check.go
```diff
@@ -0,0 +1,205 @@
...
+		ToMode:         "all",
+		ToLabelKey:     "app",
+		ToLabelValue:   "ethereum-geth-miner",
+		Duration:       dur,
+	})
+	if err != nil {
+		return err
+	}
+	log.Info().Msg("Network forked")
+	time.Sleep(dur)
```
skudasov, line 98: We are not deleting it there because experiments have finalizers of their own.

actions/setup.go
```diff
@@ -62,8 +62,11 @@ func DefaultLocalSetup(
 	if err != nil {
 		return nil, err
 	}
-	if err := contracts.AwaitMining(blockchainClient); err != nil {
-		return nil, err
+	if network.Config().Name == "Ethereum Geth reorg" {
```
jleeh, line 65: We need to never match on string names of networks. Add this in as a config flag per network, something like:
`await_mining_on_setup: true`

contracts/ethereum_reorg_util.go
```diff
@@ -27,28 +27,36 @@ func AwaitMining(c client.BlockchainClient) error {
...
+		log.Debug().Str("ExtraData", string(b.Extra())).Msg("Block received")
+		if !bytes.Contains(b.Extra(), []byte("tx")) {
```
jleeh, line 56: Shouldn't the `[]byte("tx")` here be `f.extraData`?

client/ethereum.go
```diff
@@ -411,17 +428,32 @@ func (e *EthereumClient) subscribeToNewHeaders() error {
...
+	hn := header.Number.Uint64()
+	if _, ok := e.Headers[hn]; ok {
+		e.ReorgedHeaders[hn] = header
+		return
+	}
+	e.Headers[hn] = header
+}
+
 func (e *EthereumClient) receiveHeader(header *types.Header) {
+	if header == nil {
```
jleeh, line 450: In what circumstances is this going to be nil? Same with the block.

client/ethereum.go
```diff
@@ -411,17 +428,32 @@ func (e *EthereumClient) subscribeToNewHeaders() error {
...
+			e.recordHeader(header)
 			e.receiveHeader(header)
 		case <-e.doneChan:
 			return nil
 		}
 	}
 }
 
+// recordHeader saves header either to first seen headers or to reorged/duplicated
+func (e *EthereumClient) recordHeader(header *types.Header) {
```
jleeh, line 440: I'm not a fan of always recording headers as this means memory usage will grow the longer the test runs/more blocks received. Can this be changed into a header subscription where the test will wait for a re-org and within that subscriber it records block headers?

suite/chaos/reorg_check.go
```diff
@@ -0,0 +1,205 @@
...
+)
+
+const (
+	MinerIDTemplate    = "miner-%d"
+	ReorgCheckAttempts = 15
+	ReorgCheckInterval = 10 * time.Second
+)
+
+// ReorgChecker collects data to verify reorged blocks
+type ReorgChecker struct {
```
jleeh, line 30: As per my above comment, I feel this object can be converted to a header receiver for the ETH client. It fits the pattern.

suite/chaos/reorg_check.go
```diff
@@ -0,0 +1,205 @@
...
+	AltClients []client.BlockchainClient
+	FromBlock  uint64
+	HeadersMu  *sync.Mutex
+	Headers    map[uint64]map[string]*types.Header
+	TXHashes   []common.Hash
+}
+
+// NewReorgChecker created new re-org checker connected to all miner nodes
+func NewReorgChecker(connectedClient client.BlockchainClient, env environment.Environment, cfg *config.Config, altMinersPort uint16) (*ReorgChecker, error) {
+	sd, err := env.GetAllServiceDetails(altMinersPort)
```
jleeh, line 44: This ties into multi-network support ultimately, so rather than getting all of the service details in here and creating new clients, we should alter the overall setup of the framework to handle multiple clients.

For example, in `environment.go` we should change `NewBlockchainClient` to `NewBlockchainClients` to which it inits all the available Geth clients in a cluster, meaning you don't have to do it here and just use env directly.

suite/testcommon/runlog.go
```diff
@@ -99,3 +106,23 @@ func SetupRunlogTest(i *RunlogSetupInputs) {
 		Expect(i.Err).ShouldNot(HaveOccurred())
 	})
 }
+
+func CheckRunlogCompleted(i *RunlogSetupInputs) {
+	By("receives API call data on-chain", func() {
+		err := retry.Do(func() error {
```
jleeh, line 112: Want to avoid using the retry in favour of header subscribers.

jleeh (changes requested):

client/blockchain.go
```diff
@@ -51,9 +58,12 @@ func NewBlockchainClient(network BlockchainNetwork) (BlockchainClient, error) {
 type BlockchainNetwork interface {
 	GasUsedEstimations
 	ID() string
+	SetID(id string)
```
jleeh, line 61: This method isn't used, can be removed.

client/blockchain.go
```diff
@@ -51,9 +58,12 @@ func NewBlockchainClient(network BlockchainNetwork) (BlockchainClient, error) {
...
 	URL() string
+	URLS() []string
```
jleeh, line 63: Rename to `URLs`

client/blockchain.go
```diff
@@ -51,9 +58,12 @@ func NewBlockchainClient(network BlockchainNetwork) (BlockchainClient, error) {
...
 	Type() string
 	SetURL(string)
+	SetURLS(urls []string)
```
jleeh, line 66: `SetURLs`

suite/chaos/reorg_confirmer.go
```diff
@@ -0,0 +1,242 @@
...
+		env:                 env,
+		c:                   c,
+		blockReceivers:      make([]*AggregatingBlockReceiver, 0),
+		blocksByNodeMu:      &sync.Mutex{},
+		reorgedBlocksByNode: make(map[int][]*types.Block),
+		blocksByNode:        make(map[uint64]map[int]NodeBlock),
+		blocksChan:          make(chan NodeBlock),
+	}
+	rc.subscribe()
+	rc.aggregateBlocks()
```
jleeh, line 56: `go rc.aggregateBlocks()`
Remove the go func wrapper in the function body.

suite/chaos/reorg_confirmer.go
```diff
@@ -0,0 +1,242 @@
...
+	joinBlockNumber     uint64
+	blockReceivers      []*AggregatingBlockReceiver
+	blocksByNodeMu      *sync.Mutex
+	reorgedBlocksByNode map[int][]*types.Block
+	blocksByNode        map[uint64]map[int]NodeBlock
+	blocksChan          chan NodeBlock
+}
+
+// NewReorgConfirmer creates a type that can create reorg chaos and confirm reorg has happened
+func NewReorgConfirmer(c client.BlockchainClient, env environment.Environment) (*ReorgConfirmer, error) {
```
jleeh, line 45: The `ReorgConfirmer` should be a header subscriber, not just use them. If you're using `retry` within function bodies, then something with the design is off.

Same with `awaitNetworkJoined`, you're awaiting an action based on new block headers, these sort of actions should be handled with subscribers, not retry functions.

jleeh (changes requested):

jleeh (approved):

PR #91, opened 2021-09-08
kafka

gheorghestrimtu:
Add kafka to the alerts env by adding the bitnami/kafka helm chart and changing the configurations of explorer to be able to connect to kafka through ssl. 
Also now it's possible to add more than one dependency group. This is needed when there is a chain of dependencies between groups.


environment/environment_templates.go
```diff
@@ -239,10 +242,21 @@ func NewChainlinkClusterForAlertsTesting(nodeCount int) K8sEnvSpecInit {
...
+		manifests: []K8sEnvResource{NewKafkaHelmChart()},
+	}
+
 	dependencyGroup := getBasicDependencyGroup()
-	addPostgresDbsToDependencyGroup(dependencyGroup, nodeCount)
 	addServicesForTestingAlertsToDependencyGroup(dependencyGroup, nodeCount)
-	return addNetworkManifestToDependencyGroup("basic-chainlink", dependencyGroup, chainlinkGroup)
+	addPostgresDbsToDependencyGroup(dependencyGroup, nodeCount)
+
+	dependencyGroups = append(dependencyGroups, kafkaDependecyGroup)
```
tateexon, line 257: You can get rid of the multiple appends by just adding them in via the declaration `dependencyGroups := []*K8sManifestGroup{kafkaDependecyGroup, dependencyGroup}`

environment/environment_templates.go
```diff
@@ -279,9 +293,11 @@ func NewMixedVersionChainlinkCluster(nodeCount, pastVersionsCount int) K8sEnvSpe
 		chainlinkGroup.manifests = append(chainlinkGroup.manifests, cManifest)
 	}
 
+	dependencyGroups := []*K8sManifestGroup{}
 	dependencyGroup := getBasicDependencyGroup()
 	addPostgresDbsToDependencyGroup(dependencyGroup, nodeCount)
-	return addNetworkManifestToDependencyGroup("mixed-version-chainlink", dependencyGroup, chainlinkGroup)
+	dependencyGroups = append(dependencyGroups, dependencyGroup)
```
tateexon, line 299: Another one that can be simplified: `dependencyGroups := []*K8sManifestGroup{dependencyGroup}`

environment/environment_templates.go
```diff
@@ -307,6 +323,32 @@ func NewGethReorgHelmChart() *HelmChart {
 	}
 }
 
+func NewKafkaHelmChart() *HelmChart {
```
tateexon, line 326: Public function should have a comment.

environment/environment_templates.go
```diff
@@ -351,31 +393,38 @@ func getBasicDependencyGroup() *K8sManifestGroup {
...
-func addNetworkManifestToDependencyGroup(envName string, dependencyGroup *K8sManifestGroup, chainlinkGroup *K8sManifestGroup) K8sEnvSpecInit {
+func addNetworkManifestToDependencyGroup(envName string, chainlinkGroup *K8sManifestGroup, dependencyGroups []*K8sManifestGroup) K8sEnvSpecInit {
 	return func(config *config.NetworkConfig) (string, K8sEnvSpecs) {
+		var specs K8sEnvSpecs
+		indexOfLastElementInDependencyGroups := len(dependencyGroups) - 1
 		switch config.Name {
 		case "Ethereum Geth reorg":
-			dependencyGroup.manifests = append(
-				dependencyGroup.manifests,
+			dependencyGroups[indexOfLastElementInDependencyGroups].manifests = append(
```
tateexon, line 402: You could probably make a reference to `dependencyGroups[indexOfLastElementInDependencyGroups].manifests` above the switch to use for all of these to make it more readable. Not a big deal though.

tateexon (changes requested):

tateexon (approved):

environment/environment_templates.go
```diff
@@ -311,6 +324,33 @@ func NewGethReorgHelmChart() *HelmChart {
 	}
 }
 
+// NewKafkaHelmChart creates new helm chart for kafka
+func NewKafkaHelmChart() *HelmChart {
+	valuesFilePath := filepath.Join(tools.ProjectRoot, "environment/charts/kafka/overrideValues.yaml")
+	overrideValues, err := chartutil.ReadValuesFile(valuesFilePath)
+	log.Debug().Interface("read values", overrideValues).Msg("read values")
```
jleeh, line 331: This debug log line still needed, or was it just for local debugging?

jleeh (changes requested):
Tiny comment, but looks good

environment/environment_templates.go
gheorghestrimtu: only local debugging, will remove

jleeh (approved):

PR #96, opened 2021-09-13
Add golangci-lint rules to check for comments on exported functions and variables

tateexon:
Added a golang-ci yaml file to be used by both the ci and the makefile
Fixed a subset of the issues but left issues in explorer, contracts, and contracts/ethereum directories to fix later

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ export GO111MODULE ?= on
 
 .PHONY: lint
 lint: ## run linter
-	${BIN_DIR}/golangci-lint --color=always run ./... -v --timeout=3m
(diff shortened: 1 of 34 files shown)


Makefile
```diff
@@ -4,7 +4,7 @@ export GO111MODULE ?= on
 
 .PHONY: lint
 lint: ## run linter
-	${BIN_DIR}/golangci-lint --color=always run ./... -v --timeout=3m
+	${BIN_DIR}/golangci-lint --color=always run ./... -v
```
skudasov, line 7: What is the default timeout? We had problems before in CI without it.

skudasov (approved):

jleeh (approved):

PR #98, opened 2021-09-16
otpe

gheorghestrimtu:
**Change:**
Broke `NewK8sEnvironment` into two parts. One part will connect to the cluster, the second part will deploy specs. 

**Motivation**
OTPE needs 2 urls that point to configs that contain lists of contract addresses and node addresses. In order to provide this contact addresses the contracts have to be deployed during the test. This is done through the `BlockchainClient`. The constructor of `BlockchainClient` needs an `Environment` variable that is returned by `NewK8sEnvironment`. With the previous setup you cannot invoke the `BlockchainClient` between deployments of different manifests.

**Change:**
Added mockserver helm chart
**Motivation:**
otpe needs to read configs from 2 endpoints that in prod come from weiwatchers, this endpoints need to be mocked with test info


suite/alerts/alerts_test.go
```diff
@@ -1,54 +1,24 @@
...
-			Expect(err).ShouldNot(HaveOccurred())
-			fmt.Println(chainlinkNodes[0].URL())
-
-			adapter, err = environment.GetExternalAdapter(suiteSetup.Env)
-			Expect(err).ShouldNot(HaveOccurred())
-			fmt.Println(adapter.ClusterURL())
-		})
-	})
-
+var _ = FDescribe("Alerts suite", func() {
```
skudasov, line 10: Please, remove focus.

suite/alerts/alerts_test.go
```diff
@@ -1,54 +1,24 @@
...
 	Describe("Alerts", func() {
 		It("Test 1", func() {
```
skudasov, line 12: Can we have more explicit naming for that test? Is it testing just the deployment with OPTE?

environment/templates/explorer-deployment.yml
```diff
@@ -60,16 +58,7 @@ spec:
...
-          volumeMounts:
-            - name: kafka-tls
-              mountPath: /etc/ssl/certs/kafka-0.keystore.pem
-              subPath: kafka-0.keystore.pem
-              readOnly: false
-            - name: kafka-tls
-              mountPath: /etc/ssl/certs/kafka.truststore.pem
-              subPath: kafka.truststore.pem
-              readOnly: false
+          image: 'studentcuza/explorer:choose-security-protocol'
```
skudasov, line 61: Maybe we can add explorer image as a var in our main config, repo/tag?

environment/templates/otpe-deployment.yml
```diff
@@ -0,0 +1,59 @@
...
+              value: ":1337"
+            - name: WEIWATCHERS_NODES_URL
+              value: {{ .Values.MockserverDependencyGroup.mockserver.nodesURL }}
+            - name: WEIWATCHERS_CHAINMONITOR_URL
+              value: {{ .Values.MockserverDependencyGroup.mockserver.contractsURL }}
+            - name: CHAIN_TYPE
+              value: eth
+            - name: LOG_LEVEL
+              value: debug
+          image: '804282218731.dkr.ecr.us-west-2.amazonaws.com/ocr-telemetry-prometheus-exporter:2496ee62ed6c81898ddcc42d40d3be1f9c7ac549'
```
skudasov, line 38: And here too, image/tag would be nice in config.

kalverra (changes requested):

config.yml
```diff
@@ -38,6 +38,12 @@ apps:
         limits:
             memory: 96Mi
             cpu: 50m
+    otpe:
```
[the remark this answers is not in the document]
gheorghestrimtu: Sergey suggested this change

environment/k8s_environment.go
```diff
@@ -123,26 +126,39 @@ func NewK8sEnvironment(
 	}
 	env.chaos = cc
 
+	return env, nil
+}
+
+// DeploySpecs deploys a all the specs in the array
```
kalverra, line 132: _deploys all the specs in the array_

`the array` is not exactly clear as to what we're deploying from. Maybe a better version would be something like 

> deploys all specs in the provided environment init function

>

kalverra (changes requested):

config.yml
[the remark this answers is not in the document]
kalverra: Ah yes, my bad, missed the section it was in, carry on.

kalverra (approved):

PR #102, opened 2021-09-23
Moonbeam testnet configuration

koteld:
Configuration for Moonbeam testnet


skudasov (changes requested):
Already have it under `moonbase_alpha` network, gas limit is roughly the same just the url is private, so we can close.

koteld:
Duplicates: https://github.com/smartcontractkit/integrations-framework/pull/100

PR #105, opened 2021-09-28
Simultaneous Networks

kalverra:
# See Bottom Comment for up to Date Info


gheorghestrimtu (approved):

environment/k8s_environment.go
```diff
@@ -51,7 +51,7 @@ type K8sEnvSpecs []K8sEnvResource
 
 // K8sEnvSpecInit is the initiator that will return the name of the environment and the specifications to be deployed.
 // The name of the environment returned determines the namespace.
```
tateexon, line 53: Does this comment line still make sense?

environment/environment_templates.go
```diff
@@ -292,7 +296,7 @@ func NewChainlinkCluster(nodeCount int) K8sEnvSpecInit {
 	dependencyGroup := getBasicDependencyGroup()
 	addPostgresDbsToDependencyGroup(dependencyGroup, nodeCount)
 	dependencyGroups := []*K8sManifestGroup{dependencyGroup}
-	return addNetworkManifestToDependencyGroup("basic-chainlink", chainlinkGroup, dependencyGroups)
```
tateexon, line 295: With the removal of all these envNames like basic-chainlink, is there an easy way to determine what the namespace is now? Does a multi-network deploy create multiple namespaces?

actions/setup.go
```diff
@@ -54,101 +101,182 @@ func DefaultLocalSetup(
...
+}
+
+// TearDown checks for test failure, writes logs if there is one, then tears down the test environment, based on the
+// keep_environments config value
+func (s *DefaultSuiteSetup) TearDown() func() {
+	return teardown(*s.Config, s.Env, s.Client)
+}
+
+// MultiNetworkSuiteSetup enables friendly usage of multiple networks
+type MultiNetworkSuiteSetup interface {
```
jleeh, line 138: 👍 for wrapping the suite setup in an interface. Could we get the existing single network suite setup to implement this interface (maybe renaming the interface to just `SuiteSetup`) and adding a `DefaultNetwork` function that returns the first network in the map/only network.

client/blockchain.go
```diff
@@ -107,6 +112,43 @@ func NewNetworkFromConfig(conf *config.Config) (BlockchainNetwork, error) {
...
+		for index, networkID := range networkIDs {
+			// Cannot handle multiple of the same network, throws our templating all out of whack and quickly becomes a
+			// massive headache without an overhaul of how we deal with our templating
+			if _, present := networkTracker[networkID]; present {
+				return nil, fmt.Errorf("Cannot use multiple of the same network, '%s'", networkID)
+			} else {
+				networkTracker[networkID] = true
+			}
+
+			networkConfig, err := conf.GetNetworkConfig(networkID)
```
jleeh, line 129: From this line below, it's the same code block as `NewNetworkFromConfig` bar from the map assignment. Can you just use `NewNetworkFromConfig` and then assign to the map if err is nil to not dupe code.

client/blockchain.go
```diff
@@ -107,6 +112,43 @@ func NewNetworkFromConfig(conf *config.Config) (BlockchainNetwork, error) {
 	)
 }
 
+// MultipleNetworks enables launching multiple networks for simultaneous usage in test scenarios
+func MultipleNetworks(networkIDs ...string) MultiNetworkInit {
+	return func(conf *config.Config) ([]BlockchainNetwork, error) {
+		networkTracker := make(map[string]bool)
+		networks := make([]BlockchainNetwork, len(networkIDs))
+		for index, networkID := range networkIDs {
+			// Cannot handle multiple of the same network, throws our templating all out of whack and quickly becomes a
```
jleeh, line 121: Is this a problem with templating since you're using `index` in the Chainlink deployments and the values assignment in the map for the network deployments doesn't matter if there's clobbering in the map?

We'll need to support multiple of the same network, as for CCIP the most basic test will be `ethereum_geth` and `ethereum_geth`.

jleeh (changes requested):

environment/k8s_environment.go
kalverra: No, we switched our namespace naming to the inputs in the suite setup a bit back, e.g.
```go
suiteSetup, err = actions.MultiNetworkSetup(
  "multi-network-chainlink",       // <<<<------------- Namespace initialization
  environment.NewChainlinkCluster(0),
  client.MultipleNetworks("ethereum_geth", "ethereum_hardhat"),
  tools.ProjectRoot,
)
```
The string that was returned by the `K8sEnvSpecInit` functions was never used, so I removed it on autopilot, assuming it was vestigial. But now I'm thinking about it more, it's a bit odd needing to supply the namespace at this level, so I'm going to look at why we delineated from the old ways.

client/blockchain.go
kalverra: Yes to a problem with templating, no to it being with chainlink. It was primarily an issue of the templating for the networks themselves. Looking at this freshly though, and with some of the changes around templating strategy in general I made, I think I can give this another shot.

At first, I imagined that the `ethereum_geth` <-> `ethereum_geth` would be easily replaced by something like `ethereum_geth` <-> `ethereum_hardhat`, but now I'm remembering that hardhat and ganache have some block subscription issues.

environment/k8s_environment.go
kalverra: Ah, when we split environment creation and spec deployment into separate parts, it necessitated this new style. I'd still like to condense it though.

environment/charts/mockserver-config/static/initializerJson.json
```diff
@@ -0,0 +1 @@
+[{"httpRequest":{"path":"/contracts.json"},"httpResponse":{"body":[{"contractAddress":"0xB7f8BC63BbcaD18155201308C8f3540b07f84F5e","contractVersion":4,"path":"test","status":"live"}]}},{"httpRequest":{"path":"/nodes.json"},"httpResponse":{"body":[{"id":"fe657ffb4eb718c1e1144d5c964ef8365d001a7aebb99ec04efa008562ac519d","nodeAddress":["0x4010f765f8d52d89680e1ac8e11b72ca34c172f7"]},{"id":"10ab47b894d4ee74d18955e1294632a195080ce59b34ea314101f48db9c4dee4","nodeAddress":["0x1f2713194baf824db18f9b2d481b7b25b9f1b313"]},{"id":"e48bd3392fe073e96f63d7411e0464ae53c27ffd8c3c4e303ee7a7d70ddfd754","nodeAddress":["0x81cd111e8d6d42940dc024ffa1566eeaebdf2c4c"]},{"id":"5ab4092e82d4268c1a933f7fd5b0e592d895469c3a45b4050795f9ce56acfb85","nodeAddress":["0xefd775ab82fdee8183262ddac7778670a11b188a"]},{"id":"c92e07507edc5beee53a09e29fa1415647f25e075d8a2e218b38db410ac40505","nodeAddress":["0x31e5ec154765e9e2fcd64220d4d97c8c96f31e4e"]}]}}]
```
gheorghestrimtu, line 1: This file is not needed.

actions/setup.go
```diff
@@ -23,24 +24,77 @@ const (
...
+	}, nil
+}
+
+// SuiteSetup enables common use cases, and safe handling of different blockchain networks for test scenarios
+type SuiteSetup interface {
+	Config() *config.Config
+	Environment() environment.Environment
+
+	DefaultNetwork() NetworkInfo
+	Network(networkID string) (NetworkInfo, error)
```
jleeh, line 79: I'd change this function so it works by index, the same as we do with wallets. `Network(index int)`

actions/setup.go
```diff
@@ -23,24 +24,77 @@ const (
...
+	Networks(networkID string) ([]NetworkInfo, error)
```
jleeh, line 80: Change this so it returns all networks.

config/config.go
```diff
@@ -22,15 +23,15 @@ const (
...
-	Network            string                    `mapstructure:"network" yaml:"network"`
-	Logging            *LoggingConfig            `mapstructure:"logging" yaml:"logging"`
-	Networks           map[string]*NetworkConfig `mapstructure:"networks" yaml:"networks"`
-	Retry              *RetryConfig              `mapstructure:"retry" yaml:"retry"`
-	Apps               AppConfig                 `mapstructure:"apps" yaml:"apps"`
-	Kubernetes         KubernetesConfig          `mapstructure:"kubernetes" yaml:"kubernetes"`
-	KeepEnvironments   string                    `mapstructure:"keep_environments" yaml:"keep_environments"`
-	Prometheus         *PrometheusConfig         `mapstructure:"prometheus" yaml:"prometheus"`
-	Contracts          *ContractsConfig          `mapstructure:"contracts" yaml:"contracts"`
+	Network            string                   `mapstructure:"network" yaml:"network"`
```
jleeh, line 26: We want to be able to define which networks at runtime, like we do now with a single network, so I'd change this to `Networks []string`. This way we can leave `ethereum_geth,ethereum_geth` as the default and at runtime, set the ENV:
`NETWORKS=matic_mumbai,binance_testnet`
etc

This way we don't need to define networks in the multiple networks in tests like you've got with the unit test. Then in CI, we can run the multi-network suite with a matrix in GHA that runs the tests on all network combinations.

jleeh (changes requested):
Stellar work.

PR #106, opened 2021-09-29
telemetry down alert tests

gheorghestrimtu:
Added 2 tests for the Telemetry Down alert.
1. The alert is fired if the nodes are registered on the offchainAggregator contract but have not run any jobs related to OCR
2. The alert is fired if after some successful rounds all participating chainlink nodes shut down

### DIFF
--- a/suite/alerts/alerts_test.go
+++ b/suite/alerts/alerts_test.go
@@ -1,26 +1,107 @@
 package alerts
 
 import (
+	"context"
+	"fmt"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	. "github.com/onsi/gomega/gstruct"
+	v1 "github.com/prometheus/client_golang/api/prometheus/v1"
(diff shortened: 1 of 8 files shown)


jleeh (changes requested):

jleeh (approved):

suite/alerts/alerts_test.go
```diff
@@ -1,26 +1,107 @@
...
+					model.LabelName("alertname"): Equal(model.LabelValue("Telemetry Down (infra)")),
+				}),
+				"State": Equal(v1.AlertState("firing")),
+			})))
+		})
+	})
 
-			err = i.SuiteSetup.Env.DeploySpecs(environment.PrometheusGroup())
+	AfterEach(func() {
+		By("Restoring chaos", func() {
```
tateexon, line 101: "Restoring choas" would imply you are turning the chaos back on. "Stopping chaos" would make more sense.

tateexon (approved):

PR #107, opened 2021-10-02
test ocr_telemetry_feed_message_report_req_median metric

gheorghestrimtu:
Added a test for the `ocr_telemetry_feed_message_report_req_median` metric.
Modified the `OCRSetupInputs` in `testcommon.ocr.go` to use `mockserver` for the adapter. Also, now every chainlink can be given different values to read from the "adapter", which allows for testing more scenarios.


tateexon (approved):

environment/environment_templates.go
```diff
@@ -315,7 +316,7 @@ func NewChainlinkClusterForAlertsTesting(nodeCount int) K8sEnvSpecInit {
 	}
 
 	dependencyGroup := getBasicDependencyGroup()
-	addServicesForTestingAlertsToDependencyGroup(dependencyGroup, nodeCount)
+	addServicesForTestingObservabilityToDependencyGroup(dependencyGroup, nodeCount)
```
skudasov, line 319: Do we really need a method with such a long name that only pushes to a slice? Maybe slice is enough I think

suite/observability/metrics_test.go
```diff
@@ -0,0 +1,66 @@
...
+		By("Kicking off first round")
+
+		testcommon.StartNewRound(testSetup, 1)
+
+		By("Comparing the metrics value with the expected value")
+
+		prometheus, err := environment.GetPrometheusClientFromEnv(testSetup.SuiteSetup.Env)
+		Expect(err).ShouldNot(HaveOccurred())
+
+		Eventually(func(g Gomega) int {
```
skudasov, line 46: Better to use `retry` for consistency so we don't have two libs for one purpose

skudasov (approved):

skudasov (changes requested):
`make test_ocr` panics, seems it need new mock server in setup

environment/environment_templates.go
gheorghestrimtu: removed this function

suite/observability/metrics_test.go
gheorghestrimtu: changed all retry's in tests to eventually

skudasov (approved):

PR #112, opened 2021-10-07
Integration CI on this repo

kalverra:
# CI

**New required checks**
1. unit
2. smoke

The ultimate goal is to move our full suite into the main chainlink repo, but until then, we'll try it out here. Our suite is now run for each PR to `main`, and the performance and chaos suites should run for each successful merge into `main`. We've also moved code coverage into `codecov` for better results there.

**P.S.**
This is a ton of commits, mostly due to needing to check changes in actual runs. I'm squashing all of them into a single CI commit when merging.


kalverra:
New required checks
1. unit
2. smoke

This is a ton of commits, mostly due to needing to check changes in actual runs. I'm squashing all of them into a single CI commit when merging.

PR #113, opened 2021-10-12
Per request id metrics for VRF/Runlog tests

skudasov:


suite/performance/common.go
```diff
@@ -99,3 +108,246 @@ func recordResults(b ginkgo.Benchmarker, ID string, results []time.Duration) err
...
+}
+
+// GetJobID gets internal job id
+func (n RunlogNodeData) GetJobID() string {
+	return n.JobID
+}
+
+// GetProvingKeyHash gets proving key hash for VRF
+func (n RunlogNodeData) GetProvingKeyHash() [32]byte {
+	return [32]byte{}
```
kalverra, line 288: Why is this empty? Does runlog not actually have one of these?

suite/performance/common.go
```diff
@@ -99,3 +108,246 @@ func recordResults(b ginkgo.Benchmarker, ID string, results []time.Duration) err
...
+			for k, v := range m {
+				c[contractAddr][k] = v
+			}
+		}
+	}
+}
+
+// LimitErrGroup implements the errgroup.Group interface, but limits goroutines to a
+// execute at a max throughput.
+type LimitErrGroup struct {
```
kalverra, line 326: This is more for my understanding: Why rate limit the errors?

suite/performance/runlog.go
```diff
@@ -71,14 +67,22 @@ func NewRunlogTest(
...
 		Deployer:    deployer,
 		adapter:     adapter,
-		jobMap:      RunlogJobMap{},
+		testResults: NewPerfRequestIDTestResults(),
+		jobMap:      ContractsNodesJobsMap{},
 	}
 }
 
-// RecordValues records Runlog metrics
+// RecordValues will query all of the latencies of the VRFConsumer and match them by RequestID
```
kalverra, line 75: > will query all of the latencies of the `VRFConsumer`

I'm assuming typo, this looks pretty similar to how VRF operates though, right?

suite/performance/runlog.go
```diff
@@ -171,27 +174,74 @@ func (f *RunlogTest) Run() error {
...
 	for {
 		select {
 		case <-ctx.Done():
 			log.Warn().Msg("Test finished")
+			time.Sleep(f.TestOptions.GracefulStopDuration)
+			cancelPerfEvents()
 			return nil
 		default:
-			log.Warn().Int("RoundID", i).Msg("New round")
+			log.Warn().Int("RoundID", currentRound).Msg("New round")
```
kalverra, line 195: Why does this necessitate a `WARN` log?

suite/performance/runlog.go
```diff
@@ -171,27 +174,74 @@ func (f *RunlogTest) Run() error {
...
 			if err := f.requestData(); err != nil {
 				return err
 			}
-			if err := f.waitRoundEnd(i); err != nil {
+			if err := f.waitRoundEnd(currentRound + 1); err != nil {
 				return err
 			}
-			i++
+			if f.TestOptions.NumberOfRounds != 0 && currentRound >= f.TestOptions.NumberOfRounds {
+				log.Warn().Msg("Final round is reached")
```
kalverra, line 203: Same here, this seems like a good thing?

kalverra (commented):
Left a few comments, need to spend a bit more time with the code and you to grasp some more of the big picture stuff.

suite/performance/runlog.go
skudasov: Just visually better, as a marker for long tests, can switch to INFO

suite/performance/common.go
skudasov: Just a generic goroutine group with a rate-limiting so we don't hit node hard

suite/performance/common.go
skudasov: Nope, kinda ugly thing, but I hope sometimes we can separate metrics collecting from test logic/deployment completely

skudasov: Will fix a comment though

kalverra (approved):

tateexon (approved):

PR #117, opened 2021-10-20
OCRSetupInputs multiple contracts

gheorghestrimtu:
In order to test the `Oracle not making observations on an OCR feed` alert I need to deploy more than one OCR contract with the same set of nodes

### DIFF
--- a/suite/testcommon/ocr.go
+++ b/suite/testcommon/ocr.go
@@ -50,6 +50,7 @@ func DeployOCRForEnv(i *OCRSetupInputs, envInit environment.K8sEnvSpecInit) {
 	})
 }
 
+// DeployContracts deploys and funds a certain number of offchain aggregator contracts
 func DeployContracts(i *OCRSetupInputs, nrOfOCRContracts int) {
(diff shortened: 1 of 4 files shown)


suite/testcommon/ocr.go
```diff
@@ -50,83 +50,88 @@ func DeployOCRForEnv(i *OCRSetupInputs, envInit environment.K8sEnvSpecInit) {
...
-			ethAmount,
-			big.NewFloat(2),
-		)
-		Expect(err).ShouldNot(HaveOccurred())
-
-		// Deploy and config OCR contract
-		deployer, err := contracts.NewContractDeployer(i.NetworkInfo.Client)
-		Expect(err).ShouldNot(HaveOccurred())
+// DeployContracts deploys and funds a certain number of offchain aggregator contracts
+func DeployContracts(i *OCRSetupInputs, nrOfOCRContracts int) {
```
kalverra, line 54: Might want to name this something that more explicitly points out it's for OCR only, like `DeployOCRContracts`

gheorghestrimtu: For sure

kalverra (approved):

PR #129, opened 2021-11-04
Cleanup packages, common usage and restructure

jleeh:


suite/smoke/contracts_ocr_test.go
```diff
@@ -1,23 +1,23 @@
...
 		envInit environment.K8sEnvSpecInit,
 	) {
-		i := &testcommon.OCRSetupInputs{}
-		testcommon.DeployOCRForEnv(i, envInit)
-		testcommon.FundNodes(i)
-		testcommon.DeployOCRContracts(i, 1)
-		testcommon.SendOCRJobs(i)
-		testcommon.CheckRound(i)
+		i := &actions.OCRSetupInputs{}
+		By("Deploying environment", actions.DeployOCRForEnv(i, envInit))
```
kalverra, line 16: I like this a lot more

kalverra (approved):

client/chainlink_models.go
```diff
@@ -654,3 +654,17 @@ func ObservationSourceSpecBridge(bta BridgeTypeAttributes) string {
 		parse [type=jsonparse path="data,result"];
 		fetch -> parse;`, bta.Name, bta.RequestData)
 }
+
+// marshallTemplate Helper to marshall templates
+func marshallTemplate(jobSpec interface{}, name, templateString string) (string, error) {
```
skudasov, line 659: Actually, now it's two of them

skudasov (approved):

client/chainlink_models.go
jleeh: Yeah, I didn't want to create a separate package or just a utils for a simple go templates func that'll probably never change.

PR #134, opened 2021-11-08
Change container resources

gheorghestrimtu:


kalverra (approved):
Looks good, thanks for organizing and updating.

config.yml
```diff
@@ -43,12 +49,57 @@ apps:
...
+            cpu: 1
+        limits:
+            memory: 1G
+            cpu: 1.5
     atlas-evm:
         image: "804282218731.dkr.ecr.us-west-2.amazonaws.com/atlas-evm"
         version: "2931b7e2c13ad08742a9d825c915cc01663de5c0"
+        requests:
+            memory: 200M
+            cpu: 0.25
```
tateexon, line 72: Do we want to standardize on a memory and cpu format, we are kind of using all the different possibilities here? Some of the cpu amounts are using "m" amounts whereas others are using whole numbers. Some of the memory amounts are using "Mi", some are using "M", and some are using what is potentially not even valid for memory with "m".

gheorghestrimtu: You are right we should standardize, what should we choose?
```
You can express memory as a plain integer or as a fixed-point
 number using one of these suffixes: E, P, T, G, M, k. 
You can also use the power-of-two equivalents: 
Ei, Pi, Ti, Gi, Mi, Ki. For example, the following represent
 roughly the same value:

128974848, 129e6, 129M, 123Mi
```

config.yml
```diff
@@ -43,12 +49,57 @@ apps:
...
+        limits:
+            memory: 350m
+            cpu: 0.4
+    kafka-rest:
+        image: confluentinc/cp-kafka-rest
+        version: 7.0.0
+        requests:
+            memory: 150m
```
tateexon, line 80: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#resource-units-in-kubernetes
You can express memory as a plain integer or as a fixed-point number using one of these suffixes: E, P, T, G, M, k. You can also use the power-of-two equivalents: Ei, Pi, Ti, Gi, Mi, Ki

So thus these lowercase 'm' should be invalid.

tateexon (changes requested):

config.yml
gheorghestrimtu: This are the choices for cpu

The expression 0.1 is equivalent to the expression 100m, which can be read as "one hundred millicpu". Some people say "one hundred millicores", and this is understood to mean the same thing. A request with a decimal point, like 0.1, is converted to 100m by the API, and precision finer than 1m is not allowed. For this reason, the form 100m might be preferred.

config.yml
gheorghestrimtu: I guess it doesn't care about capitalization, I deployed with this code and it worked

config.yml
tateexon: In a lot of the infra stuff I have been using for EAs they have been using 'm' for cpu and the Ei, Pi, Ti, Gi, Mi, Ki options for memory. If we go that route we would be consistent with them.

tateexon: I am not partial to any particular way though, so long as we are consistent.

tateexon (approved):

PR #138, opened 2021-11-09
Remove dummy adapter

gheorghestrimtu:


suite/smoke/multi_network_contracts_test.go
```diff
@@ -29,7 +29,7 @@ var _ = Describe("Basic Contract Interactions @contract", func() {
 		By("Deploying the environment", func() {
 			var err error
 			suiteSetup, err = actions.MultiNetworkSetup(
-				environment.NewChainlinkCluster(0),
+				environment.NewChainlinkCluster(1),
```
kalverra, line 32: Is this necessary? This test only makes sure that contracts are launched, with no interaction with chainlink nodes.

suite/smoke/contracts_test.go
```diff
@@ -25,7 +25,7 @@ var _ = Describe("Basic Contract Interactions @contract", func() {
 		By("Deploying the environment", func() {
 			var err error
 			suiteSetup, err = actions.SingleNetworkSetup(
-				environment.NewChainlinkCluster(0),
+				environment.NewChainlinkCluster(1),
```
kalverra, line 28: Same here.

suite/performance/vrf_soak_test.go
```diff
@@ -68,7 +65,6 @@ var _ = Describe("VRF soak test @soak-vrf", func() {
 				defaultNetwork.Client,
 				defaultNetwork.Wallets,
 				defaultNetwork.Deployer,
-				adapter,
```
kalverra, line 71: Was this just not being used in the VRF tests?

kalverra (changes requested):

suite/smoke/multi_network_contracts_test.go
gheorghestrimtu: If it's zero, the test fails with 
E1109 22:03:11.641977   12154 portforward.go:400] an error occurred forwarding 45191 -> 9417: error forwarding port 9417 to pod 30a423d36711bfd5d1f1110b27de89c052d3e0101c81463cee30c141443a2b31, uid : exit status 1: 2021/11/09 22:03:11 socat[16410] E connect(5, AF=2 127.0.0.1:9417, 16): Connection refused

suite/performance/vrf_soak_test.go
gheorghestrimtu: yes

client/mockserver.go
```diff
@@ -36,6 +36,27 @@ func (em *MockserverClient) ClearExpectation(body interface{}) error {
 	return err
 }
 
+// SetVariable sets an int for /variable
+func (em *MockserverClient) SetVariable(v int) error {
```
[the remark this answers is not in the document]
gheorghestrimtu: But it has more usages besides as a mock adapter, for instance in o11y tests I used to mock weiwatchers

suite/smoke/multi_network_contracts_test.go
gheorghestrimtu: I think it has something to do with this change
https://github.com/smartcontractkit/integrations-framework/pull/138/files#diff-dedb54e5426283c5c6cefbe9a1587207f8e8060396acd9e2914b6194630f61f7R679

But didn't dig deeper to find out why

gheorghestrimtu: Also hopefully the env changes that @skudasov is making will fix this

kalverra: Wouldn't it make sense to just move the mock server manifests into the `basicDependencyGroup`?

gheorghestrimtu: good idea

skudasov (approved):

skudasov (approved):

kalverra (approved):

PR #150, opened 2021-11-24
Update OCR actions

gheorghestrimtu:


client/blockchain.go
```diff
@@ -188,28 +188,30 @@ func (n *NetworkRegistry) GetNetworks(env *environment.Environment) (*Networks,
 }
 
 // NewChainlinkClients creates new chainlink clients
-func NewChainlinkClients(e *environment.Environment) ([]Chainlink, error) {
+func NewChainlinkClients(e *environment.Environment, charts []string) ([]Chainlink, error) {
```
jleeh, line 191: To avoid changing all the tests to always need to pass in a string slice of what 9/10 times will be the same, create two functions, eg:

```go
func NewChainlinkClients(e *environment.Environment) ([]Chainlink, error) {
  return NewChainlinkClientsByCharts(e, []string{"chainlink"})
}
```

suite/smoke/ocr_test.go
```diff
@@ -2,167 +2,77 @@ package smoke
...
 				environment.NewChainlinkConfig(environment.ChainlinkReplicas(6, nil)),
 				tools.ChartsRoot,
 			)
 			Expect(err).ShouldNot(HaveOccurred())
-			err = e.ConnectAll()
+			err = env.ConnectAll()
 			Expect(err).ShouldNot(HaveOccurred())
 		})
-
 		By("Getting the clients", func() {
```
jleeh, line 28: I prefer the pattern:
```go
By("Getting the clients", actions.NewOCRSetup(env))
```
You could keep the functions in the actions the same, but wrap them to handle error assertions, eg:

```go
func GinkgoWrap(fn func() error) func() {
  return func() {
    Expect(fn()).ShouldNot(HaveOccurred())
  }
}
```

actions/ocr.go
```diff
@@ -1,293 +1,202 @@
...
+	uuid "github.com/satori/go.uuid"
+	"github.com/smartcontractkit/helmenv/environment"
+	"github.com/smartcontractkit/integrations-framework/client"
+	"github.com/smartcontractkit/integrations-framework/contracts"
+	"math/big"
+	"time"
+)
+
+// OCRSetup contains the components needed for testing OCR
+type OCRSetup struct {
```
jleeh, line 15: I know this is from the older implementation, but I want to avoid needing custom structs for each action type, as it blocks composability of action functions between different types. It'd be much better to always pass the parameters in individually to the action functions.

jleeh (changes requested):

jleeh (approved):

kalverra (approved):

PR #174, opened 2021-12-31
Improvemets: test JSON RPC compatibility

koteld:


openrpc.json
```diff
@@ -0,0 +1,4302 @@
+{
+	"openrpc": "1.2.4",
```
kalverra: Can we move this to live in the folder with the tests? And maybe run it through a minimizer?

suite/compatibility/json_rpc_test.go
```diff
@@ -0,0 +1,140 @@
...
+			}
+		})
+	})
+
+	Describe("Test GET RPC methods and validate results", func() {
+		It("OCR test GET Methods", func() {
+			for chainId, rpcClients := range rpcClientsByChain {
+				fmt.Printf("Starting tests for chain ID %d\n", chainId)
+				for rpcMethod, rpcMethodParameters := range rpcMethodCalls {
+					fmt.Printf("\n\nMethod: %s", rpcMethod)
```
kalverra, line 98: Make use of `zerolog` instead of the `fmt.Print`s.

suite/compatibility/json_rpc_test.go
```diff
@@ -0,0 +1,140 @@
...
+		}},
+	}
+
+	BeforeEach(func() {
+		By("Getting RPC clients", func() {
+			nc, err := config.LoadNetworksConfig(filepath.Join(utils.ProjectRoot, "networks.yaml"))
+			Expect(err).ShouldNot(HaveOccurred())
+			for _, networkName := range nc.SelectedNetworks {
+				networkSettings, ok := nc.NetworkSettings[networkName]
+				Equal(ok)
```
kalverra, line 79: It doesn't seem clear what this is checking or if it's really asserting anything. Does `Equal` run an assert on `ok` this way? Doesn't it need to have something to check what it's equal against?

suite/compatibility/json_rpc_test.go
```diff
@@ -0,0 +1,140 @@
...
+						if value.Name == rpcMethod {
+							method = value
+							break
+						}
+					}
+					schemaLoader := gojsonschema.NewGoLoader(method.Result.Schema)
+					for _, rpcClient := range rpcClients {
+						var rpcCallResult interface{}
+						err := rpcClient.CallContext(context.Background(), &rpcCallResult, rpcMethod, rpcMethodParameters...)
+						if err != nil {
```
kalverra, line 110: Make use of the Ginkgo matchers and assertions instead.

kalverra (changes requested):

openrpc.json
koteld: Yes, for sure, we can do that.
Also, we can try compiling this file from source just before the test: https://github.com/ethereum/execution-apis

suite/compatibility/json_rpc_test.go
koteld: It's like assert.True(...), but I'll try to make it more obvious, or will remove this assertion.

kalverra: Ah okay, that's fine then. I wasn't aware of that behavior.

suite/compatibility/json_rpc_test.go
koteld: I want test to continue so I don't use gomega assertions and use logging instead.
I'm not sure if it is reasonable to make separate tests for each JSON RPC method and use assertion for them.
If it is possible to continue test after failed assertion I will use it.

PR #177, opened 2022-01-06
Bug/keeper fix

kalverra:
# Fixes for Keeper and Expandability

This PR is sizeable, so let me break down some of the larger points.

## Contract Recompilation

Keeper contracts had some small tweaks that I included and then recompiled. To do so, I pulled out a python script that got lost in the shuffle before. It can be converted to Go in the future no problem if we like, but it's pretty trivial. A lot of the file changes are just recompiling our contracts with this new Keeper Contract Update

## Configurability

I added in a couple easy framework config settings so you can change chainlink and get images, versions, and env vars/CLI args

## Unit Tests

Added some more of these


tateexon (approved):
lgtm

PR #196, opened 2022-01-24
support adding terra chains

RyanRHall:


jmank88 (approved):

go.mod
```diff
@@ -4,6 +4,7 @@ go 1.17
 
 require (
 	github.com/ethereum/go-ethereum v1.10.11
+	github.com/guregu/null v3.5.0+incompatible
```
jmank88, line 7: Should this be `gopkg.in/guregu/null.v4`?

jmank88 (approved):

go.mod
RyanRHall: yeah idk why it keeps trying to make me use 3.5.0...

RyanRHall: oh I see, it's the wrong package. `gopkg.in/guregu/null` vs `github.com/guregu/null`

jmank88 (approved):

skudasov (approved):

PR #214, opened 2022-02-09
sqlx PG connector

skudasov:


client/chainlink.go
```diff
@@ -73,13 +73,20 @@ type Chainlink interface {
 
 	// SetClient is used for testing
 	SetClient(client *http.Client)
+
+	DB() *PostgresConnector
```
jleeh, line 77: Remove the DB out of the Chainlink client and have it as its own client, it shouldn't be a prerequisite to using the client as it could break downstream implementations.

jleeh (changes requested):

jleeh (approved):

tateexon (approved):

PR #222, opened 2022-02-16
It's My Money, And I Want It... When You're Done With It

kalverra:
This PR makes returning funds from chainlink nodes part of the tear-down step. The logic was surprisingly complex due to how chainlink manages transactions, but I managed to get a fairly performant process going. Hopefully, I can work with devs to institute an easier way, but until then, this should do the job well.

The added `returnFunds(chainlinkNodes []client.Chainlink, networks *client.Networks) error` function and its many helpers do the following:

1. Loop through each network
2. If that network is a simulated network (e.g. local geth instance) we don't bother to return the funds
3. Otherwise we go to all chainlink nodes, and if their addresses have a balance, attempt to send the remaining balance back to the network's default wallet address. (The gas calculation is a bit dirty on this, no good solution at the moment)
4. The chainlink nodes are monitored over 100 seconds to check if the transactions actually confirmed. On confirmation, we shut everything down like normal.
5. If any of these steps fail, we print out a warning "We couldn't get your money back. Leaving the environment running so you can try to do it manually." and then... well then we do that.

This doesn't make tests fail if the attempts to send back funds fail. Tests will still pass, but you'll get a big mean log message telling you what went wrong.

I also threw in a QoL framework setting that allows you to write logs whether your test passes or not. I found this to be handy for debugging while working out the quirks on this one.


testsetups/keeper.go
```diff
@@ -94,8 +94,6 @@ func (k *KeeperBlockTimeTest) Setup() {
 		Expect(err).ShouldNot(HaveOccurred(), "Deploying KeeperConsumerPerformance instance %d shouldn't fail", i+1)
 		k.keeperConsumerContracts = append(k.keeperConsumerContracts, keeperConsumerInstance)
 		Expect(err).ShouldNot(HaveOccurred())
-		// err = k.linkTokenContract.Transfer(keeperConsumerInstance.Address(), big.NewInt(1e18))
```
tateexon, line 97: Do we still need these commented out lines?

kalverra: Nope, they were an unnecessary step that I took out as an experiment to see if they were truly unnecessary. I had forgotten to fully remove them, doing so now.

actions/actions.go
```diff
@@ -203,10 +203,148 @@ func TeardownSuite(
...
+		// Send async request to each chainlink node to send a transaction back to the network default wallet
+		sendFundsErrGroup.Go(
+			func() error {
+				primaryEthKeyData, err := node.ReadPrimaryETHKey()
+				if err != nil {
+					return err
+				}
+				nodeBalanceString := primaryEthKeyData.Attributes.ETHBalance
+				if nodeBalanceString != "0" { // If key has a non-zero balance, attempt to transfer it back
+					gasCost, err := network.EstimateTransactionGasCost()
```
skudasov, line 270: Why do we care about gas costs here at all? Isn't standard gas estimation for EVM compatible clients is enough?

actions/actions.go
```diff
@@ -203,10 +203,148 @@ func TeardownSuite(
...
+						return confirmTransaction(node, sentFromAddress, toAddress, transactionErrGroup)
+					})
+				} else {
+					log.Debug().Int("Node Number", nodeIndex).Msg("Chainlink node had no funds to return")
+				}
+			}
+
+			return transactionErrGroup.Wait()
+		},
+		retry.Delay(time.Second*5),
```
skudasov, line 317: We have two retry libraries, again, one from `ginkgo` and that one

client/chainlink.go
```diff
@@ -298,23 +299,44 @@ func (c *chainlink) ReadETHKeys() (*ETHKeys, error) {
 	return ethKeys, err
 }
 
-// CreateOCR2Key creates an OCR2Key on the Chainlink node
+// ReadPrimaryETHKey reads updated information about the chainlink's primary ETH key
+func (c *chainlink) ReadPrimaryETHKey() (*ETHKeyData, error) {
```
skudasov, line 303: If that is a private key why can't we do that without node API?

suite/smoke/keeper_test.go
```diff
@@ -179,7 +172,7 @@ var _ = Describe("Keeper suite @keeper", func() {
 			networks.Default.GasStats().PrintStats()
 		})
 		By("Tearing down the environment", func() {
-			err = actions.TeardownSuite(env, networks, utils.ProjectRoot, nil)
+			err = actions.TeardownSuite(env, networks, utils.ProjectRoot, chainlinkNodes, nil)
```
skudasov, line 175: That will break all the stuff in Terra/Solana on update, can we start to use functional options there?

actions/actions.go
kalverra: I'm assuming you're referring to the Ginkgo `Eventually`? We can switch to that, but it seemed out of place here, as this shouldn't be asserting anything or failing any tests.

suite/smoke/keeper_test.go
kalverra: Messaging you

tateexon (approved):

skudasov (approved):

PR #230, opened 2022-03-02
Remote Test Runner

kalverra:
Hooks into the new `remote-test-runner` in the new version of `helmenv`, enabling you to launch long-running soak tests from the `suite/soak/soak_runner_test.go` file which looks like the below:

```go
	t.Parallel()
	actions.LoadConfigs(utils.ProjectRoot)
	log.Logger = log.Output(zerolog.ConsoleWriter{Out: os.Stderr})
	exePath, remoteConfig := buildGoTests(t)

	env, err := environment.DeployLongTestEnvironment(
		environment.NewChainlinkConfig(environment.ChainlinkReplicas(6, nil), "chainlink-soak"),
		tools.ChartsRoot,
		remoteConfig.TestRegex,                             // Name of the test to run
		remoteConfig.SlackWebhookURL,                       // Slack Webhook to hit when test finished
		remoteConfig.SlackAPIKey,                           // API key to use to upload artifacts to slack
		filepath.Join(utils.ProjectRoot, "framework.yaml"), // Path of the framework config
		filepath.Join(utils.ProjectRoot, "networks.yaml"),  // Path to the networks config
		exePath, // Path to the executable test file
	)
	require.NoError(t, err)
	require.NotNil(t, env)
	log.Info().Str("Namespace", env.Namespace).
		Str("Environment File", fmt.Sprintf("%s.%s", env.Namespace, "yaml")).
		Msg("Soak Test Successfully Launched. Save the environment file to collect logs when test is done.")
```


testreporters/ocr.go
```diff
@@ -0,0 +1,268 @@
...
+		}
+	}
+	ocrReportWriter.Flush()
+
+	log.Info().Str("Location", filepath.Join(utils.ProjectRoot, "ocr_soak_report.csv")).Msg("Wrote CSV file")
+	return nil
+}
+
+// sends a slack message to a slack webhook if one is provided
+func (o *OCRSoakTestReporter) sendSlack(csvLocation, namespace string) error {
```
tateexon, line 141: Do we want this function to be moved to a more common part of the library since there is nothing in here specific to ocr and it would make sending to slack easier for other test setups?

kalverra: Ya, was recently thinking about that since it would be helpful in the keeper tests we want to run.We can tack it on to this PR, or the next one that will move the performance tests to also use the remote runner.

tateexon (approved):

client/blockchain.go
```diff
@@ -254,6 +283,29 @@ func ConnectChainlinkNodesByCharts(e *environment.Environment, charts []string)
...
 
+// ConnectChainlinkNodesSoak assumes that the tests are being run from an internal soak test runner
+func ConnectChainlinkNodesSoak(e *environment.Environment) ([]Chainlink, error) {
+	var clients []Chainlink
+
+	remoteURLs, err := e.Charts.Connections("chainlink").RemoteURLsByPort("access", environment.HTTP)
+	if err != nil {
+		return nil, err
+	}
+	for urlIndex, localURL := range remoteURLs {
```
gheorghestrimtu, line 294: I think here should be
```
for urlIndex, remoteURL := range remoteURLs
```

testsetups/ocr.go
```diff
@@ -0,0 +1,210 @@
...
+}
+
+// checks on all OCR instances that they all received the correct answer from the latest round
+func (t *OCRSoakTest) checkLatestRound(expectedValue, roundNumber int) {
+	var roundAnswerGroup sync.WaitGroup
+	roundAnswerChannel := make(chan latestRoundAnswer, len(t.ocrInstances))
+	for _, ocrInstance := range t.ocrInstances {
+		roundAnswerGroup.Add(1)
+		ocrInstance := ocrInstance
+		go func() {
```
gheorghestrimtu, line 178: here maybe you could do something like
```
    done := make(chan interface{})
    go func() {
      defer GinkgoRecover()

      content := book.RawContent()
      Expect(book.Pages).To(Equal(2783))

      //this might be quite expensive and will block...
      err := book.SetFontSize(28)
      Expect(err).NotTo(HaveOccurred())

      Expect(book.Pages).To(BeNumerically(">", 2783))
      Expect(book.RawContent()).To(Equal(content))

      close(done)
    }()

    Eventually(done).Should(BeClosed())
```

from https://onsi.github.io/ginkgo/#patterns-for-asynchronous-testing

kalverra: The other defer is to control the wait group - assuming there is no failure - which I just pushed a bit of a cleaner version.

gheorghestrimtu (approved):

PR #244, opened 2022-03-28
Move fcdUrl from node attributes to chain attributes

gheorghestrimtu:


gheorghestrimtu:
the relevant issue
https://github.com/smartcontractkit/chainlink-terra/issues/78

PR #257, opened 2022-04-06
Keeper block time

kalverra:
* Updates Keeper Performance Contracts
* Adds a Keeper Block Time Test
* Makes configuring simulated Geth and Chainlink Nodes easier
* Updates docs


gheorghestrimtu (approved):

actions/ocr_helpers.go
```diff
@@ -24,25 +24,59 @@ func DeployOCRContracts(
...
+		Expect(err).ShouldNot(HaveOccurred(), "Error getting node's primary ETH address")
+		transmitters = append(transmitters, addr)
+		payees = append(payees, networks.Default.GetDefaultWallet().Address())
+	}
+
+	// Set Payees
+	for contractCount, ocrInstance := range ocrInstances {
+		err = ocrInstance.SetPayees(transmitters, payees)
+		Expect(err).ShouldNot(HaveOccurred(), "Error setting OCR payees")
+		if (contractCount+1)%100 == 0 { // For large amounts of contract deployments, space things out some
```
tateexon, line 57: Would you want to break this 100 out to a variable since it is consistent in 3 places and you likely would want to update all 3 to be the same if it needs to be updated in the future?

tateexon (approved):

PR #268, opened 2022-04-14
Klatyn Fixes

kalverra:
Fixes some quirks of [Katlyn](https://www.klaytn.com/) and its odd [gas calculation](https://docs.klaytn.com/klaytn/design/transaction-fees#unit-price) that entirely breaks down when met with our way of bumping gas. Word on the street is they're changing this "soon".

Thanks and credit to @koteld on finding and fixing an intermittent bug on long-running OCR tests.


client/ethereum.go
```diff
@@ -525,7 +526,16 @@ func (e *EthereumClient) SendTransaction(
 	if err != nil {
 		return common.Hash{}, err
 	}
+
+	// https://docs.klaytn.com/klaytn/design/transaction-fees#unit-price
```
jleeh, line 530: Create a `KlaytnClient` that inherits the ETH client and override these methods so it doesn't dirty up the main client with string matching

jleeh (changes requested):

client/ethereum.go
kalverra: I was considering this, but since it's such a tiny difference, figured it was maybe not worth. But ya, less string matching == better.

PR #275, opened 2022-04-20
fund 0x0 when using Geth by default

skudasov:


suite/smoke/keeper_test.go
```diff
@@ -61,7 +61,7 @@ var _ = Describe("Keeper suite @keeper", func() {
 			err = actions.FundChainlinkNodes(chainlinkNodes, networks.Default, txCost)
 			Expect(err).ShouldNot(HaveOccurred(), "Funding Chainlink nodes shouldn't fail")
 			// Edge case where simulated networks need some funds at the 0x0 address in order for keeper reads to work
-			if networks.Default.GetNetworkType() == "eth_simulated" {
+			if networks.Default.GetNetworkType() == client.SimulatedEthNetwork {
```
jleeh, line 64: Is this still needed in both of these tests with it being done already in fund?

jleeh (changes requested):

jleeh (approved):

PR #299, opened 2022-05-11
Remote Test Controller

kalverra:
Adds a basic endpoint that can be called on a soak test to shut it down gracefully. Sometimes we get into situations where we need to kill a soak test before it can complete, but would still like to get info from it; this enables such a thing.

You can access an endpoint on the `remote-test-runner`, `/shutdown` that will ask you to confirm you want a shut down with a random code. Calling `/shutdown/<code>` will gracefully shutdown the test and give you a truncated report.

### DIFF
--- a/suite/soak/soak_runner_test.go
+++ b/suite/soak/soak_runner_test.go
@@ -25,7 +25,7 @@ func TestSoak(t *testing.T) {
 	env, err := environment.DeployLongTestEnvironment(
 		environment.NewChainlinkConfig(
 			environment.ChainlinkReplicas(6, config.ChainlinkVals()),
-			"chainlink-soak",
+			"shutdown-test",
 			config.GethNetworks()...,
 		),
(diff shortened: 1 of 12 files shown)


suite/soak/soak_runner_test.go
```diff
@@ -25,7 +25,7 @@ func TestSoak(t *testing.T) {
 	env, err := environment.DeployLongTestEnvironment(
 		environment.NewChainlinkConfig(
 			environment.ChainlinkReplicas(6, config.ChainlinkVals()),
-			"chainlink-soak",
+			"shutdown-test",
```
tateexon, line 28: Did you mean to change this?

kalverra: Only temporarily

tateexon (approved):

skudasov (approved):

PR #311, opened 2022-05-18
Add keeper registry 1.2 contract and interface

infiloop2:
In this PR I'm adding the new keeper registry contract (v1.2) and also supporting the old one
Did a bit of refactoring to pass the parameters. The interface for the contract remains same, but some functions have different implementation depending on the registry version

Corresponding CL repo PR: CL repo PR: https://github.com/smartcontractkit/chainlink/pull/6646
where both registry versions are tested in smoke tests

**Note**: I ran `make compile_contracts` and it also updated the other wrappers with latest abigen


contracts/contract_models.go
```diff
@@ -186,123 +186,6 @@ type MockGasFeed interface {
...
-		upkeepAddr string,
-		gasLimit uint32,
-		adminAddr string,
-		checkData []byte,
-		amount *big.Int,
-		source uint8,
-	) ([]byte, error)
-	Fund(ethAmount *big.Float) error
-}
-
```
infiloop2, line 210: consolidated these all in contracts/ethereum_keeper_contracts.go

contracts/contract_models.go
```diff
@@ -186,123 +186,6 @@ type MockGasFeed interface {
 	Address() string
 }
 
-type UpkeepRegistrar interface {
```
infiloop2, line 189: consolidated these all in contracts/ethereum_keeper_contracts.go

.gitignore
```diff
@@ -27,8 +27,8 @@ artifacts/
 # JUnit Test Reports
 junit.xml
 tests-*.xml
-logs/
-bin/
+/logs/
+/bin/
```
infiloop2, line 31: Before this was ignoring the bin generated for contracts, but I saw that there are already files committed for those so clarified this for consistency

tateexon (approved):
looks good to me

PR #312, opened 2022-05-19
Removed observation source injection from Keepers-related tests

vla33:
Because the observation source is the same for all Keeper jobs (as part of [this](https://github.com/smartcontractkit/chainlink/pull/6582) PR in the main repo), the users do not need to pass it anymore in a Toml string when creating the jobs.

### DIFF
--- a/actions/keeper_helpers.go
+++ b/actions/keeper_helpers.go
@@ -39,7 +39,6 @@ func CreateKeeperJobs(chainlinkNodes []client.Chainlink, keeperRegistry contract
 			ContractAddress:          keeperRegistry.Address(),
 			FromAddress:              chainlinkNodeAddress,
 			MinIncomingConfirmations: 1,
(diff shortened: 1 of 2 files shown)


actions/keeper_helpers.go
```diff
@@ -39,7 +39,6 @@ func CreateKeeperJobs(chainlinkNodes []client.Chainlink, keeperRegistry contract
 			ContractAddress:          keeperRegistry.Address(),
 			FromAddress:              chainlinkNodeAddress,
 			MinIncomingConfirmations: 1,
-			ObservationSource:        client.ObservationSourceKeeperDefault(),
```
kalverra, line 42: We can probably rip out this function entirely.

vla33: Thanks for pointing it out Adam! Just removed it now.

infiloop2 (approved):

tateexon (approved):

PR #317, opened 2022-05-23
refactor to chainlink-env, new configs and connection strategy for clients

skudasov:
Example of a `chainlink-env` environment, ready for review.
Migrated all tests in core to that new version.
Removed all the tests, except that we don't have in the core repo, refactored connections, and simplified `Networks`


actions/soak_helpers.go
```diff
@@ -5,18 +5,26 @@ import (
...
-	"github.com/smartcontractkit/helmenv/environment"
 )
 
 // Soak Test helpers
 
 // BuildGoTests builds the go tests to run, and returns a path to it, along with remote config options
-func BuildGoTests(executablePath, testsPath string) (string, error) {
-	exePath := filepath.Join(executablePath, "remote.test")
+func BuildGoTests(testsPath string) (string, string, error) {
+	exePath := filepath.Join(utils.ProjectRoot, "remote.test")
```
kalverra, line 27: This could lead to issues when running soak tests from other repos. I saw this a while back where we would run from `test-repo/suite` and the soak tests would be looking for an executable in `github.com/smartcontractkit/chainlink-testing-framework@v1/...` and fail. Worth checking running from an outside repo.

actions/soak_helpers.go
```diff
@@ -28,47 +36,75 @@ func BuildGoTests(executablePath, testsPath string) (string, error) {
...
-	)
+	env := environment.New(&environment.Config{
+		Labels:    []string{fmt.Sprintf("envType=%s", pkg.EnvTypeEVM5RemoteRunner)},
+		Namespace: namespacePrefix,
+	})
+	err = env.
+		AddHelm(mockservercfg.New(nil)).
+		AddHelm(mockserver.New(nil)).
+		AddHelm(remotetestrunner.New(map[string]interface{}{
+			"remote_test_runner": map[string]interface{}{
```
kalverra, line 64: See my concerns with this chart style not being intuitive. This invites a whole lot of user errors inputting the wrong keys and values. Maybe best replaced with a struct.

blockchain/ethereum.go
```diff
@@ -716,18 +717,23 @@ func (e *EthereumMultinodeClient) WaitForEvents() error {
 }
 
 // SimulatedEthereumURLs returns the websocket URLs for a simulated geth network
-func SimulatedEthereumURLs(e *environment.Environment) ([]*url.URL, error) {
-	return e.Charts.Connections("geth").LocalURLsByPort("ws-rpc", environment.WS)
+func SimulatedEthereumURLs(e *environment.Environment) ([]string, error) {
```
kalverra, line 720: Can we combine `SimulatedEthereumURLs` and `SimulatedSoakEthereumURLs` to just `SimulatedEthereumURLs` that first looks for a "local" forwarded port, and if not there, will try and find a "remote" one?

config/config.go
```diff
@@ -55,25 +54,25 @@ func ChainlinkVals() map[string]interface{} {
...
-			gethNetworks = append(gethNetworks, environment.DefaultGeth)
-		case PerformanceGeth:
-			gethNetworks = append(gethNetworks, environment.PerformanceGeth)
-		case RealisticGeth:
-			gethNetworks = append(gethNetworks, environment.RealisticGeth)
-		}
-	}
-	return gethNetworks
-}
+//// GethNetworks builds the proper geth network settings to use based on the selected_networks config
```
kalverra, line 57: So we're no longer going to pick up different geth types to launch?

suite/performance/basic_test.go
```diff
@@ -5,53 +5,55 @@ import (
...
-					config.ChainlinkVals(),
-					"chainlink-profiling",
-					config.GethNetworks()...,
-				),
-			)
-			Expect(err).ShouldNot(HaveOccurred(), "Environment deployment shouldn't fail")
-			err = testEnvironment.ConnectAll()
+			e = environment.New(nil)
+			err := e.
+				AddHelm(mockservercfg.New(nil)).
```
kalverra, line 38: `AddDeployment()` instead of `AddHelm()`? K8s has proven to be enough of a stumbling block for most outside users, would rather not imply they need to know about helm as well.

suite/performance/basic_test.go
```diff
@@ -5,53 +5,55 @@ import (
...
+				AddHelm(mockserver.New(nil)).
+				AddHelm(geth.New(nil)).
```
kalverra, line 40: Will this always launch a simulated geth instance? Regardless if we want to only run on some testnet like Kovan?

suite/soak/tests/ocr_test.go
```diff
@@ -22,11 +27,17 @@ var _ = Describe("OCR Soak Test @soak-ocr", func() {
 
 	BeforeEach(func() {
 		By("Deploying the environment", func() {
-			env, err = environment.DeployOrLoadEnvironmentFromConfigFile(
-				"/root/test-env.json", // Default location for the soak-test-runner container
+			env = environment.New(
+				&environment.Config{InsideK8s: true, TTL: 12 * time.Hour},
```
kalverra, line 31: Assuming `TTL` will tear down the environment after time limit?

kalverra (changes requested):

suite/soak/tests/ocr_test.go
skudasov: Set default TTL to 168 hours for soak

suite/performance/basic_test.go
skudasov: There is a multinetwork example I'm working on, which will be updated, but for now, we have limitation of only one networks

suite/performance/basic_test.go
skudasov: Looking to simplify that, but for now, it's better to explicitly state which kind of deployment we are using, `chainlink-env` have multiple.

config/config.go
skudasov: See comment about multinetwork preset

blockchain/ethereum.go
skudasov: Nice catch, fixed, moved paths up

actions/soak_helpers.go
skudasov: Nice catch, moved the paths up

actions/soak_helpers.go
skudasov: Nothing I can do about it now if we add a struct it needs to be transformed into a map, so another layer of abstraction, we don't need, idea is that we do not enforce structure on an arbitrary charts that may be external

client/chainlink.go
```diff
@@ -700,93 +703,53 @@ func (c *chainlink) do(
...
-	return ConnectChainlinkDBByCharts(e, []string{"chainlink"})
-}
-
-// ConnectChainlinkDBByCharts creates new chainlink DBs clients by charts
-func ConnectChainlinkDBByCharts(e *environment.Environment, charts []string) ([]*PostgresConnector, error) {
 	var dbs []*PostgresConnector
-	for _, chart := range charts {
-		pgUrls, err := e.Charts.Connections(chart).LocalURLsByPort("postgres", environment.HTTP)
+	dbURLs := e.URLs[chainlink2.DBsLocalURLsKey]
+	for _, u := range dbURLs {
```
kalverra, line 713: Could use some more descriptive variable names, not really sure what you're doing here, or why.

config/test_framework_config.yml
```diff
@@ -5,28 +5,4 @@
 keep_environments: Never # Options: Always, OnFail, Never
 logging:
     # panic=5, fatal=4, error=3, warn=2, info=1, debug=0, trace=-1
-    level: 5
-
-# Specify the image and version of the chainlink image you want to run tests against. Leave blank for default.
-chainlink_image: testChainlinkImage
```
kalverra, line 11: Why remove these test values?

suite/framework.yaml
```diff
@@ -6,15 +6,4 @@ keep_environments: Never # Options: Always, OnFail, Never
 logging:
     write_pod_logs: OnFail # Options: Always, OnFail
     # panic=5, fatal=4, error=3, warn=2, info=1, debug=0, trace=-1
-    level: 0
-
-# Specify the image and version of the chainlink image you want to run tests against. Leave blank for default.
-chainlink_image: public.ecr.aws/chainlink/chainlink
```
kalverra, line 12: How do we control which version and values we're working with? Only through ENV vals?

suite/smoke/vrf_test.go
```diff
@@ -36,21 +41,17 @@ var _ = Describe("VRF suite @vrf", func() {
...
-				environment.NewChainlinkConfig(
-					config.ChainlinkVals(),
-					"chainlink-vrf",
-					config.GethNetworks()...,
-				),
-			)
-			Expect(err).ShouldNot(HaveOccurred(), "Environment deployment shouldn't fail")
-			err = e.ConnectAll()
-			Expect(err).ShouldNot(HaveOccurred(), "Connecting to all nodes shouldn't fail")
+			e = environment.New(nil).
```
kalverra, line 44: Is there any way to provide namespace names?

Are simulated geth instances always launched by default now?

kalverra (changes requested):

suite/framework.yaml
skudasov: Yes, why do we need a config and env vars at the same time?

suite/smoke/vrf_test.go
skudasov: Sure, here is the config for now
// Config environment configuration
type Config struct {
	TTL               time.Duration
	NamespacePrefix   string
	Namespace         string
	Labels            []string
	NSLabels          *map[string]*string
	ReadyCheckData    *client.ReadyCheckData
	DryRun            bool
	InsideK8s         bool
	KeepConnection    bool
	RemoveOnInterrupt bool
}

skudasov: Yes, we should just use table tests to inject anything you want and then run that particular case, like:
image

suite/framework.yaml
skudasov: Also, I removed it from here because I'd like to have deployment vars and test harness vars decoupled, `chainlink_image` has nothing to do with a framework, not only CL node can be deployed.

suite/smoke/vrf_test.go
skudasov: We can also tear down `Networks` and `networks.yaml` because in fact, when we need different clients + deployments ( in external network case I still would like to have no-op deployment just to contain URLs and params too ) and it's a single network case - we can use table test an just inject both client with config and deployment you want, making that's explicit just adds to readability

If CL requires multiple networks, that'll be a completely different suite.

skudasov: Because when we have one client - we don't need `Networks` wrapper, when we have 1+ we don't need `DefaultNetwork`, just make client calls explicitly, make things simpler

client/chainlink.go
skudasov: Sure, fixed.

actions/soak_helpers.go
```diff
@@ -5,57 +5,38 @@ import (
...
 	"github.com/smartcontractkit/chainlink-testing-framework/config"
 	"github.com/smartcontractkit/chainlink-testing-framework/utils"
-	"github.com/smartcontractkit/helmenv/environment"
 )
 
-// BuildGoTestsDocker builds the go tests to run using docker, and returns a path to the test executable, along with
-// remote config options. This version usually takes longer to run, but eliminates issues with cross-compilation.
-//  Note: currentProjectRootPath and currentSoakTestRootPath are not interchangeable with utils.ProjectRoot and utils.SoakRoot
-//  when running in outside repositories. Keep an eye on when you need paths leading to this go package vs the current running project.
-func BuildGoTestsDocker(currentProjectRootPath, currentSoakTestRootPath, testsPath string) (string, error) {
```
kalverra, line 19: We'll still want these to help get around execution and compilation errors.

suite/reorg/direcct_request.go
```diff
@@ -1,67 +1,105 @@
-package smoke
+package reorg
```
kalverra, line 1: Why not move this to core with the others?

testsetups/common.go
```diff
@@ -0,0 +1,28 @@
+package testsetups
```
kalverra, line 1: Maybe more appropriate in the `blockchain` package?

testsetups/ocr.go
```diff
@@ -30,8 +31,7 @@ type OCRSoakTest struct {
 
 	env            *environment.Environment
 	chainlinkNodes []client.Chainlink
-	networks       *blockchain.Networks
-	defaultNetwork blockchain.EVMClient
+	c              blockchain.EVMClient
```
kalverra, line 34: Need a better variable name than `c`

kalverra (changes requested):

actions/vrfv2_helpers.go
```diff
@@ -2,7 +2,6 @@ package actions
```
tateexon, line 2: Most of this file just needs changes reverted and should just have your client/networks changes.

suite/soak/soak_runner_test.go
```diff
@@ -1,28 +1,33 @@
 package soak_test
```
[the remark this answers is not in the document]
skudasov: Moving all the tests that are left in a separate PR, just need to validate for several repos, core, and atlas, so let's put it out of scope.

suite/reorg/direcct_request.go
skudasov: The same as for soak tests.

testsetups/profile.go
```diff
@@ -17,9 +18,8 @@ type ChainlinkProfileTest struct {
 	Inputs       ChainlinkProfileTestInputs
 	TestReporter testreporters.ChainlinkProfileTestReporter
 
-	env            *environment.Environment
-	networks       *blockchain.Networks
-	defaultNetwork blockchain.EVMClient
+	env *environment.Environment
+	c   blockchain.EVMClient
```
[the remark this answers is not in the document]
skudasov: Collapses with `client` because we have the same package, I'm personally okay if we'll have `c` `e` and `cl` because we'll always have them, or wrap them together soon.

blockchain/blockchain.go
```diff
@@ -4,39 +4,21 @@ package blockchain
...
 	"github.com/smartcontractkit/chainlink-testing-framework/config"
 )
 
 // Commonly used blockchain network types
 const (
-	SimulatedEthNetwork   = "eth_simulated"
-	LiveEthTestNetwork    = "eth_testnet"
-	LiveKlaytnTestNetwork = "klaytn_testnet"
-	LiveMetisTestNetwork  = "metis_testnet"
+	SimulatedEthNetwork = "eth_simulated"
```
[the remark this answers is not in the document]
skudasov: Yeah, for returning funds, we still need to differentiate simulated vs external.

actions/soak_helpers.go
skudasov: We'll do that separately when moving soak tests.

kalverra (approved):

PR #320, opened 2022-05-24
Add helper functions for keepers

infiloop2:
- Return upkeep IDs after registration. These are fetched from logs resulting from link.transferAndCall tx
- Add function to set config on registry

Corresponding CL repo PR here:  https://github.com/smartcontractkit/chainlink/pull/6673

### DIFF
--- a/contracts/ethereum_contracts.go
+++ b/contracts/ethereum_contracts.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
 	"github.com/rs/zerolog/log"
 	"github.com/smartcontractkit/chainlink-testing-framework/blockchain"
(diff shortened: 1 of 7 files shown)


kalverra (approved):

contracts/ethereum_contracts.go
```diff
@@ -550,22 +551,23 @@ func (l *EthereumLinkToken) Transfer(to string, amount *big.Int) error {
 	return l.client.ProcessTransaction(tx)
 }
 
-func (l *EthereumLinkToken) TransferAndCall(to string, amount *big.Int, data []byte) error {
+func (l *EthereumLinkToken) TransferAndCall(to string, amount *big.Int, data []byte) (*types.Transaction, error) {
```
vla33, line 554: Nice one! 😊 I think it's much better to have the `TransferAndCall` method have the same return type as the methods belonging to the `LinkToken` struct to avoid confusion.

infiloop2: I think this change actually makes the return value different than other functions :sweat_smile:, but other functions haven't needed the tx as output yet.

PR #321, opened 2022-05-24
Vrfv2 soak test functionality

tateexon:


actions/vrfv2_helpers.go
```diff
@@ -0,0 +1,84 @@
...
+	err = networks.Default.WaitForEvents()
+	Expect(err).ShouldNot(HaveOccurred())
+
+	return coordinator, consumer, bhs
+}
+
+func CreateVrfV2Jobs(
+	chainlinkNodes []client.Chainlink,
+	coordinator contracts.VRFCoordinatorV2,
+) ([]*client.Job, [][2]*big.Int) {
```
kalverra, line 37: Any way we can get a bit more type-safety around the proving keys? Not a big deal, just feels like an oddly unique data type to just throw out there.

actions/vrfv2_helpers.go
```diff
@@ -0,0 +1,84 @@
...
+		}
+		ost, err := os.String()
+		Expect(err).ShouldNot(HaveOccurred())
+		oracleAddr, err := n.PrimaryEthAddress()
+		Expect(err).ShouldNot(HaveOccurred())
+		job, err := n.CreateJob(&client.VRFV2JobSpec{
+			Name:                     fmt.Sprintf("vrf-%s", jobUUID),
+			CoordinatorAddress:       coordinator.Address(),
+			FromAddress:              oracleAddr,
+			EVMChainID:               "1337",
```
kalverra, line 57: This should pull from the Network config's chain ID

actions/vrfv2_helpers.go
```diff
@@ -0,0 +1,84 @@
...
+		provingKey := Vrfv2RegisterProvingKey(vrfKey, oracleAddr, coordinator)
+		encodedProvingKeys = append(encodedProvingKeys, provingKey)
+	}
+	return jobs, encodedProvingKeys
+}
+
+func Vrfv2RegisterProvingKey(
+	vrfKey *client.VRFKey,
+	oracleAddress string,
+	coordinator contracts.VRFCoordinatorV2) [2]*big.Int {
```
kalverra, line 75: Formatting... I think. It might be legal Go, but it looks odd to me. The below would fit conventions more I think.

```go
func Vrfv2RegisterProvingKey(
	vrfKey *client.VRFKey,
	oracleAddress string,
	coordinator contracts.VRFCoordinatorV2,
) [2]*big.Int {
```

actions/vrfv2_helpers.go
tateexon: It is odd looking indeed. Will do.

testsetups/vrfv2.go
```diff
@@ -0,0 +1,169 @@
...
+	// set the requests to only run for a certain amount of time
+	testContext, testCancel := context.WithTimeout(context.Background(), t.Inputs.TestDuration)
+	defer testCancel()
+
+	t.NumberOfRequests = 0
+
+	// variables dealing with how often to tick and how to stop the ticker
+	stop := false
+	startTime := time.Now()
+	ticker := time.NewTicker(time.Minute / time.Duration(t.Inputs.RequestsPerMinute))
```
kalverra, line 113: This might be easier with a context? [I did OCR with one](https://github.com/smartcontractkit/chainlink-testing-framework/blob/main/testsetups/ocr.go#L127) and was pleased, but it might not work for your specific case. Maybe worth a look.

kalverra (changes requested):

testsetups/vrfv2.go
tateexon: Your implementation doesn't tick on the same interval every time. It is ticking every (time it takes to make all the calls + TimeBetweenRounds) which is too limiting for some of the higher requests per minute the vrf team want.

kalverra (approved):

actions/vrfv2_helpers.go
```diff
@@ -0,0 +1,88 @@
...
+
+	. "github.com/onsi/gomega"
+	"github.com/rs/zerolog/log"
+	uuid "github.com/satori/go.uuid"
+	"github.com/smartcontractkit/chainlink-testing-framework/blockchain"
+	"github.com/smartcontractkit/chainlink-testing-framework/client"
+	"github.com/smartcontractkit/chainlink-testing-framework/contracts"
+)
+
+func DeployVrfv2Contracts(
```
jleeh, line 16: Nitpick, VRF is capitalised elsewhere but not in these funcs

jleeh (changes requested):

jleeh (approved):

PR #322, opened 2022-05-25
Add more functionality for keeper contracts

infiloop2:
KeeperPerformanceContracts
- Return upkeep IDs when these are deployed
- Expose setCheckGasToBurn, setPerformGasToBurn functions

KeeperRegistry
- SetGasLimit for upkeep

Keeper Helpers:
- Pass in upkeepGasLimit as arg
- Set the default wallet as admin

These functions are used in CL PR:  https://github.com/smartcontractkit/chainlink/pull/6678


suite/smoke/keeper_test.go
```diff
@@ -81,6 +81,7 @@ var _ = Describe("Keeper suite @keeper", func() {
 					FallbackLinkPrice:    big.NewInt(2e18),
 				},
 				1,
+				uint32(2500000), //upkeepGasLimit
```
vla33, line 84: I'm just wondering, is there a maximum gas limit that a user can set when registering their upkeep? If there is, it might be worth making it a constant and exposing it to other packages.

infiloop2: It's dependent on the maxPerformGas set in registry config which can be set differently in every test

RyanRHall (approved):

suite/smoke/keeper_test.go
vla33: I see, thanks!

vla33 (approved):

PR #353, opened 2022-06-22
fix a nil address issue

tateexon:
I must have moved code around when doing cleanup because using l.address before assigning it in doesn't work.


kalverra (approved):

AnieeG:
I see this commit is already available in `main`? am I missing something here?

PR #367, opened 2022-06-29
Story/39197 : contract event listener

AnieeG:
Listen to the events published by ocr contracts and validate the current answer whenever there is an "AnswerUpdated" event gets generated. Additionally the other event details can also be captured in logs ( commented code in `ocr.go`), if it's needed.


testsetups/ocr.go
```diff
@@ -230,6 +238,73 @@ func (t *OCRSoakTest) checkLatestRound(expectedValue, roundNumber int) {
...
+			select {
+			case err := <-sub.Err():
+				Expect(err).ShouldNot(HaveOccurred(), "Retrieving event subscription log in OCR instances shouldn't fail")
+				continue
+			case vLog := <-eventLogs:
+				// the first topic is the hashed event signature
+				eventDetails, err := contractABI.EventByID(vLog.Topics[0])
+				Expect(err).ShouldNot(HaveOccurred(), "Getting event details for OCR instances shouldn't fail")
+				// The following is a generic details of any event received.
+				/*parsedEvent, err := contractABI.Unpack(eventDetails.Name, vLog.Data)
```
tateexon, line 271: Do we still need this or is it something that is just useful to keep commented?

AnieeG: Actually looking for suggestions here. There are multiple events published. If we think we might need to capture all such event details, I can remove the comments. Otherwise I can get rid of this snippet.
This actually corresponds to this part of pr description:

> 
Additionally the other event details can also be captured in logs ( commented code in ocr.go), if it's needed.

kalverra (approved):

testsetups/ocr.go
tateexon: I guess I am not sure what all is in the event off the top of my head. It might be worthwhile keeping the comment in for anyone that wants to tap into it for a quick test in the future.

testsetups/ocr.go
```diff
@@ -230,6 +238,67 @@ func (t *OCRSoakTest) checkLatestRound(expectedValue, roundNumber int) {
...
+		query.Addresses = append(query.Addresses, common.HexToAddress(t.ocrInstances[i].Address()))
+	}
+	eventLogs := make(chan types.Log)
+	sub, err := t.chainClient.SubscribeFilterLogs(context.Background(), query, eventLogs)
+	if err != nil {
+		Expect(err).ShouldNot(HaveOccurred(), "Subscribing to contract event log in OCR instance shouldn't fail")
+	}
+	ocr := t.ocrInstances[0]
+	go func() {
+		//defer GinkgoRecover()
```
tateexon, line 259: Do we need this? I know GinkgoRecover can be really wonky in these goroutines sometimes so I am fine with leaving the comment here.

tateexon (approved):

PR #368, opened 2022-06-30
Small Teaks

kalverra:


tateexon (approved):

testsetups/keeper.go
```diff
@@ -26,7 +26,6 @@ type KeeperBlockTimeTest struct {
 
 	keeperRegistry          contracts.KeeperRegistry
 	keeperConsumerContracts []contracts.KeeperConsumerPerformance
-	mockServer              *client.MockserverClient
```
AnieeG, line 29: is it moved to some other structure or not needed at all?

kalverra: Not needed, seems the test doesn't utilize it at all

AnieeG (approved):

PR #381, opened 2022-07-14
Story/39608 test refactor CL client

AnieeG:
Corresponding PR in core - https://github.com/smartcontractkit/chainlink/pull/6975


client/chainlink_models.go
```diff
@@ -1,25 +1,33 @@
...
+	Type       string                `json:"type"`
+	ID         string                `json:"id"`
+	Attributes TransactionAttributes `json:"attributes"`
+}
 
-// EIServiceConfig represents External Initiator service config
-type EIServiceConfig struct {
-	URL string
+type TransactionAttributes struct {
+	State    string `json:"state"`
```
skudasov, line 15: TransactionAttributes/TransactionsData/TransactionData/TransactionsMetaData can be moved too because they are ETH specific.

client/chainlink_models.go
```diff
@@ -1,25 +1,33 @@
 package client
```
skudasov, line 1: Please, also move `basic_http_client.go` and remove `client_suite.go`. We won't need them after refactoring with `resty`.

skudasov (changes requested):
Nice work, Ani! Added some nitpicks, but LTGM!

client/chainlink_models.go
AnieeG: `basic_http_client.go` is being used by other clients like kafka, mockserver. I can remove `client_suite.go`

client/chainlink_models.go
AnieeG: Discussed offline. These are all used in testreports.

skudasov (approved):

kalverra (approved):

PR #388, opened 2022-07-25
Add keepers benchmark test

anirudhwarrier:
Add keeper benchmark test contracts


anirudhwarrier:
Related PR in chainlink repo - https://github.com/smartcontractkit/chainlink/pull/7066

contracts/ethereum/v0.7/src/KeeperConsumerBenchmark.sol
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  function setSpread(uint256 _newTestRange, uint256 _newAverageEligibilityCadence) external {
+    testRange = _newTestRange;
+    averageEligibilityCadence = _newAverageEligibilityCadence;
+  }
+
+  function rand() private view returns (uint256) {
+    return uint256(keccak256(abi.encode(blockhash(block.number - 1), address(this))));
+  }
```
infiloop2, line 85: nit. this function doesn't seem to be used

contracts/ethereum/v0.7/src/KeeperConsumerBenchmark.sol
```diff
@@ -0,0 +1,86 @@
...
+    return eligible();
+  }
+
+  function reset() external {
+    initialCall = 0;
+    count = 0;
+  }
+
+  function setSpread(uint256 _newTestRange, uint256 _newAverageEligibilityCadence) external {
+    testRange = _newTestRange;
```
infiloop2, line 79: we should set `initialCall` = 0 here? Not sure if we are using this in the test, if not then i would say just remove it

contracts/ethereum/v0.7/src/KeeperConsumerBenchmark.sol
```diff
@@ -0,0 +1,86 @@
...
+    checkGasToBurn = _checkGasToBurn;
+    performGasToBurn = _performGasToBurn;
+  }
+
+  function checkUpkeep(bytes calldata data) external view returns (bool, bytes memory) {
+    uint256 startGas = gasleft();
+    bytes32 dummyIndex = blockhash(block.number - 1);
+    bool dummy;
+    // burn gas
+    while (startGas - gasleft() < checkGasToBurn) {
```
infiloop2, line 28: I'm thinking whether we should add some buffer here to ensure checkUpkeep doesn't take more than checkGasToBurn? (e.g. https://github.com/smartcontractkit/chainlink/blob/develop/contracts/src/v0.7/tests/UpkeepMock.sol#L12-L32) 

Not a strong opinion but could be helpful in debugging

infiloop2 (changes requested):
Added some inline comments

contracts/ethereum/v0.7/src/KeeperConsumerBenchmark.sol
anirudhwarrier: These were copied from https://github.com/smartcontractkit/chainlink/blob/develop/contracts/src/v0.7/tests/UpkeepPerformCounterRestrictive.sol.
I think the idea behind this setter is to be able to modify the cadence of upkeep on the fly. There is a separate reset function to reset the counter. atm this not being used anywhere in the test. I'd prefer it stays here in case a future version of the test needs it.

contracts/ethereum/v0.7/src/KeeperConsumerBenchmark.sol
anirudhwarrier: removed

contracts/ethereum/v0.7/src/KeeperConsumerBenchmark.sol
anirudhwarrier: Since this was working fine for tests done until now, I think it is best to not make this change. Moreover, any changes would mean the test results would be invalidated.

infiloop2 (approved):

gheorghestrimtu (approved):

PR #397, opened 2022-07-31
Confirm Events

kalverra:


blockchain/ethereum.go
```diff
@@ -343,6 +343,81 @@ func (e *EthereumClient) ProcessTransaction(tx *types.Transaction) error {
...
+	eventName string,
+	event *types.Log,
+	eventConfirmedChan chan bool,
+	eventErrorChan chan error,
+) {
+	minConfirmations := e.NetworkConfig.MinimumConfirmations
+
+	go func() {
+		headerChannel := make(chan *types.Header)
+		subscription, err := e.Client.SubscribeNewHead(context.Background(), headerChannel)
```
skudasov, line 358: We are already subscribing each client to react on a new block in `receiveHeader` and `sub.ReceiveBlock`, can we do that according to our reactive mode in the same style as `NewTransactionConfirmer`
There are also more clear ways to share all blocks by a node in multinode client without that reactive style, but before we have time to refactor I suppose a unified approach is preferred?

skudasov (changes requested):

blockchain/ethereum.go
kalverra: > can we do that according to our reactive composition in the same style as `NewTransactionConfirmer`

I tinkered around with this but abandoned it as it got even messier than this version. Though I'm on board to do it. Might be I just need to return with a fresh set of eyes.

tateexon (approved):

skudasov (approved):

PR #402, opened 2022-08-03
Added workDir support for Gauntlet commands

smickovskid:
- Added workDir support for Gauntlet commands


skudasov (approved):

client/evm.go
```diff
@@ -0,0 +1,53 @@
...
+type Node struct {
+	ID        int32     `json:"ID"`
+	Name      string    `json:"Name"`
+	ChainID   string    `json:"ChainID"`
+	URL       string    `json:"URL"`
+	CreatedAt time.Time `json:"CreatedAt"`
+	UpdatedAt time.Time `json:"UpdatedAt"`
+}
+
+func CreateNodeKeysBundle(nodes []*client.Chainlink, chainName string) ([]NodeKeysBundle, error) {
```
krebernisak, line 25: Where was this code moved from?

It probably fits more to move it as part of the Chainlink client in the core repo `"github.com/smartcontractkit/chainlink/integration-tests/client"` (next to the code it uses)

gauntlet/gauntlet.go
```diff
@@ -16,6 +16,10 @@ import (
 	"github.com/rs/zerolog/log"
 )
 
+var (
```
krebernisak, line 19: @RodrigoAD we should find a better location for the Gauntlet Go wrapper, any ideas? 

Maybe https://github.com/smartcontractkit/gauntlet-core

krebernisak (approved):

client/evm.go
smickovskid: This was code that was adjusted based on how keys were created in Solana, we had no methods for EVM based chains so idea was to move it here since the setup will always be same for evm chains

PR #405, opened 2022-08-04
Instant to L2 txs

kalverra:
Instant confirmations have been causing some intermittent issues, especially with L2 chains. This new strategy still has txs confirmed quickly, without using only No-Ops.


tateexon (approved):

blockchain/ethereum.go
```diff
@@ -328,7 +328,7 @@ func (e *EthereumClient) TransactionOpts(from *EthereumWallet) (*bind.TransactOp
 func (e *EthereumClient) ProcessTransaction(tx *types.Transaction) error {
 	var txConfirmer HeaderEventSubscription
 	if e.GetNetworkConfig().MinimumConfirmations == 0 {
-		txConfirmer = &InstantConfirmations{}
+		txConfirmer = NewL2TxConfirmer(e, tx.Hash())
```
skudasov, line 331: How come we have `NewL2TxConfirmer` if `e.GetNetworkConfig().MinimumConfirmations == 0`, is all networks with zero confirmations required are L2?

blockchain/ethereum.go
```diff
@@ -347,7 +347,7 @@ func (e *EthereumClient) ProcessTransaction(tx *types.Transaction) error {
 func (e *EthereumClient) ProcessEvent(name string, event *types.Log, confirmedChan chan bool, errorChan chan error) error {
 	var eventConfirmer HeaderEventSubscription
 	if e.GetNetworkConfig().MinimumConfirmations == 0 {
-		eventConfirmer = &InstantConfirmations{}
+		eventConfirmer = NewL2TxConfirmer(e, event.TxHash)
 	} else {
 		eventConfirmer = NewEventConfirmer(name, e, event, e.GetNetworkConfig().MinimumConfirmations, errorChan, confirmedChan)
```
skudasov, line 352: What is the logic behind having both TxConfirmer and EventConfirmer determined because of `MinimumConfirmations`? What that parameter then stands for? Tx confirmations or events?

blockchain/transaction_confirmers.go
```diff
@@ -102,22 +101,71 @@ func (t *TransactionConfirmer) Complete() bool {
...
+		l.complete = true
+	}()
+	if l.complete {
+		return nil
+	}
+
+	for {
+		select {
+		case <-countdown.C:
+			confirmed, err := l.client.IsTxConfirmed(l.txHash)
```
skudasov, line 151: We have confirmations here on ticks and in `ReceiveBlock` one is not enough?

skudasov (changes requested):

blockchain/ethereum.go
kalverra: They all essentially behave as Optimistic L2s, with near-instant confirmations. We shouldn't assume instant confirmation, this gets us in trouble with race conditions between our tests and the chain. (e.g. Instant confirming deployment of an OCR contract that we then try to configure a few ms before it exists on chain)

blockchain/ethereum.go
kalverra: We can tweak that, but I think the idea is that for anything on-chain to count as "real", it has to pass this minimum confirmation count.

blockchain/transaction_confirmers.go
kalverra: Faster this way. On Arbitrum testing, this cut down wait times by ~1s on average.

skudasov: Well, that is really, really strange, because I get that L2 block time is fast, but then they should be confirmed in the next block at least, not before.

PR #406, opened 2022-08-05
Re-Subscribe To New Blocks on Error

kalverra:


blockchain/transaction_confirmers.go
```diff
@@ -262,7 +262,12 @@ func (e *EthereumClient) subscribeToNewHeaders() error {
 	for {
 		select {
 		case err := <-subscription.Err():
-			return err
+			log.Error().Err(err).Msg("Error while subscribed to new headers, restarting subscription")
+
+			subscription, err = e.Client.SubscribeNewHead(context.Background(), headerChannel)
+			if err != nil {
+				return err
+			}
```
skudasov, line 270: `err` returned from `e.Client.SubscribeNewHead` is not the same that can return from `<-subscription.Err()` so that won't loop, even if it will do there is no timeout

skudasov (changes requested):

blockchain/transaction_confirmers.go
skudasov: Maybe it will, but I can't be sure, can we test it first?

kalverra: Would love to test it, but it seems the only way to do so is to let it soak for 1+ weeks to hit the error we see on longer subscriptions. So will need to just wait for that to come through, or find a way to trigger the error on a sub.

skudasov (approved):

blockchain/transaction_confirmers.go
kalverra: > err returned from e.Client.SubscribeNewHead is not the same that can return from <-subscription.Err() so that won't loop

Right, the error returned from an initial sub is usually something along the lines of the RPC client not being up anymore at all. We'd want some more comprehensive looping and error handling around errors like that.

AnieeG (approved):

PR #418, opened 2022-08-29
update workflows and tool-versions

tateexon:
Brings our .golangci file more up to date with the chainlink repo as well

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -22,11 +22,6 @@ linters:
     - misspell
     - errorlint
     - contextcheck
-  disable:
-    - deadcode
-    - rowserrcheck
-    - structcheck
-    - varcheck
 linters-settings:
   exhaustive:
     default-signifies-exhaustive: true
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -6,16 +6,13 @@ run:
 linters:
   enable:
     # defaults
-    - deadcode
     - errcheck
     - gosimple
     - govet
     - ineffassign
     - staticcheck
-    - structcheck
     - typecheck
     - unused
-    - varcheck
     # extra
     - exhaustive
     - exportloopref
@@ -23,9 +20,13 @@ linters:
     - goimports
     - gosec
     - misspell
-    - rowserrcheck
     - errorlint
     - contextcheck
+  disable:
+    - deadcode
+    - rowserrcheck
+    - structcheck
+    - varcheck
 linters-settings:
   exhaustive:
     default-signifies-exhaustive: true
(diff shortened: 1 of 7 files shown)


kalverra (approved):

.golangci.yaml
```diff
@@ -6,26 +6,27 @@ run:
...
     - gosec
     - misspell
-    - rowserrcheck
     - errorlint
     - contextcheck
+  disable:
+    - deadcode
+    - rowserrcheck
+    - structcheck
+    - varcheck
```
AnieeG, line 29: is it already being taken care by other checks and that's why disabled?

testsetups/remote_control.go
```diff
@@ -41,6 +41,7 @@ func StartRemoteControlServer(name string, stopChan chan struct{}) {
 
 	// Start listening for shutdown calls
 	go func() {
+		//nolint:gosec
```
AnieeG, line 44: if not needed can this be removed?

gauntlet/gauntlet.go
```diff
@@ -68,7 +68,8 @@ type ExecCommandOptions struct {
 }
 
 // ExecCommand Executes a gauntlet command with the provided arguments.
-//  It will also check for any errors you specify in the output via the errHandling slice.
+//
```
AnieeG, line 71:
```suggestion
```

actions/soak_helpers.go
```diff
@@ -16,8 +16,9 @@ import (
 
 // BuildGoTests builds the go tests using native go cross-compilation to run, and returns a path to the test executable
 // along with its size in bytes.
-//  Note: currentProjectRootPath and currentSoakTestRootPath are not interchangeable with utils.ProjectRoot and utils.SoakRoot
-//  when running in outside repositories. Keep an eye on when you need paths leading to this go package vs the current running project.
+//
```
AnieeG, line 19:
```suggestion
```

.golangci.yaml
kalverra: They're deprecated. I don't think golangci runs them anymore though, so I think this is redundant?

tateexon: It is deprecated: https://github.com/smartcontractkit/chainlink-testing-framework/runs/8079127193?check_suite_focus=true#step:4:33

AnieeG: can it be removed then?

gauntlet/gauntlet.go
tateexon: go fmt automatically adds this and if it isn't set like this then the linter gets angry.

actions/soak_helpers.go
tateexon: go fmt automatically adds this and if it isn't set like this then the linter gets angry.

testsetups/remote_control.go
tateexon: This issue has been open since 2016 and still has no resolution: https://github.com/golang/go/issues/16100
So I skipped it with the linter

.golangci.yaml
tateexon: I will see if removing them from the disabled list does indeed not run them.

tateexon: Removing works and pr is updated

AnieeG (approved):

PR #435, opened 2022-09-13
Add http url in config

AnieeG:


blockchain/config.go
```diff
@@ -71,6 +73,7 @@ func (e *EVMNetwork) ToMap() map[string]interface{} {
 		"evm_name":                        e.Name,
 		"evm_chain_id":                    fmt.Sprint(e.ChainID),
 		"evm_urls":                        strings.Join(e.URLs, ","),
+		"evm_https_urls":                  strings.Join(e.HTTPURLs, ","),
```
kalverra, line 76: Are they `HTTP` or `HTTPS`?

AnieeG: I am dealing with simulated geth so it's http as of now. For testnets is it generally https?

kalverra: It can be both, so I think changing them all to just be `http` makes most sense

kalverra (approved):

tateexon (approved):

PR #471, opened 2022-10-27
bump chainlink-env version

tateexon:


go.mod
```diff
@@ -17,12 +17,29 @@ require (
...
 	github.com/rs/zerolog v1.28.0
 	github.com/slack-go/slack v0.11.3
-	github.com/smartcontractkit/chainlink-env v0.2.43
+	github.com/smartcontractkit/chainlink-env v0.2.45
 	github.com/stretchr/testify v1.8.1
 	go.uber.org/ratelimit v0.2.0
 	golang.org/x/sync v0.0.0-20220819030929-7fc1605a5dde
 )
 
+require (
```
kalverra, line 26: Can consolidate this block with the below. I know it's caused issues for core in the past to have these separated blocks.

kalverra (changes requested):

go.mod
tateexon: Has anybody looked into why go mod tidy creates this if is is problematic?

kalverra (approved):

skudasov (approved):

PR #510, opened 2022-12-13
Enable creating concurrent evmclient for same network

AnieeG:
Details in - https://github.com/smartcontractkit/chainlink-testing-framework/pull/509

### DIFF
--- a/blockchain/ethereum.go
+++ b/blockchain/ethereum.go
@@ -44,6 +44,11 @@ type EthereumClient struct {
 	doneChan            chan struct{}
 }
 
+type NonceSettings struct {
+	NonceMu *sync.Mutex
+	Nonces  map[string]uint64
+}
+
 // newEVMClient creates an EVM client for a single node/URL
 func newEVMClient(networkSettings EVMNetwork) (EVMClient, error) {
 	log.Info().
@@ -67,6 +72,7 @@ func newEVMClient(networkSettings EVMNetwork) (EVMClient, error) {
 		queueTransactions:   false,
 		doneChan:            make(chan struct{}),
 	}
+
 	if err := ec.LoadWallets(networkSettings); err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


blockchain/ethereum.go
```diff
@@ -76,6 +82,14 @@ func newEVMClient(networkSettings EVMNetwork) (EVMClient, error) {
 	return wrapSingleClient(networkSettings, ec), nil
 }
 
+// SyncNonce sets the NonceMu and Nonces value based on existing EVMClient
+// it ensures the instance of EthereumClient is synced with passed EVMClient's nonce updates.
+func (e *EthereumClient) SyncNonce(c EVMClient) {
+	n := c.GetNonceSetting()
```
kalverra, line 88: Do we need to utilize the NonceMu here?

blockchain/ethereum.go
```diff
@@ -646,6 +677,42 @@ func NewEVMClient(networkSettings EVMNetwork, env *environment.Environment) (EVM
...
+		}
+	}
+
+	return wrappedClient, nil
+}
+
+// ConcurrentEVMClient returns a multi-node EVM client connected to the specified network
+// It is used for concurrent interactions from different threads with the same network and from same owner
+// account. This ensures that correct nonce value is fetched when an instance of EVMClient is initiated using this method.
+func ConcurrentEVMClient(networkSettings EVMNetwork, env *environment.Environment, existing EVMClient) (EVMClient, error) {
```
kalverra, line 702: At least the beginning setup seems quite similar to `NewEVMClient`. Any way we can consolidate? Maybe we don't even need the full new method?

kalverra (changes requested):

kalverra (approved):

tateexon (approved):

blockchain/ethereum.go
AnieeG: Not changing this now as discussed. We are going to explore more robust method of handling concurrent tx from the same client

PR #516, opened 2023-01-18
Faster Live Testnets

kalverra:
## Overview

This PR enables us to run multiple tests in parallel on the same live test network. This is done by creating a global nonce-manager in `blockchain/nonce_settings.go` whenever we find that the test network is not simulated.

## Details

If the test is running on a simulated network, there are no changes.

If the test is running on a non-simulated network, a global nonce-tracker is started. From there, each client (usually 1 per test) requests nonces from this manager. This enables tracking nonces across multiple threads on the same instance.

## Caveats

Things get complicated when running on instant confirmation networks (L2s like Optimism). These networks have little to no mempools, and so we need to not only get nonces in a proper order, but also submit transactions in proper order. This requires a couple channels to inform when each transaction is good to be released.

### DIFF
--- a/blockchain/ethereum.go
+++ b/blockchain/ethereum.go
@@ -28,22 +28,7 @@ import (
 	"github.com/smartcontractkit/chainlink-testing-framework/utils"
 )
 
-// Used for when running tests on a live test network, so tests can share nonces and run in parallel
-var globalNonceManager = newNonceSettings()
-
-// convenience function
-func newNonceSettings() *NonceSettings {
-	return &NonceSettings{
-		NonceMu: &sync.Mutex{},
-		Nonces:  make(map[string]uint64),
-	}
-}
-
-// NonceSettings is a convenient wrapper for holding nonce state
-type NonceSettings struct {
-	NonceMu *sync.Mutex
-	Nonces  map[string]uint64
-}
+var debugID int
 
 // EthereumClient wraps the client and the BlockChain network to interact with an EVM based Blockchain
 type EthereumClient struct {
@@ -58,6 +43,7 @@ type EthereumClient struct {
 	queueTransactions   bool
 	gasStats            *GasStats
 	doneChan            chan struct{}
+	debugID             int
 }
 
 // newEVMClient creates an EVM client for a single node/URL
@@ -72,8 +58,10 @@ func newEVMClient(networkSettings EVMNetwork) (EVMClient, error) {
 		return nil, err
 	}
 
+	debugID++
 	ec := &EthereumClient{
 		NetworkConfig:       networkSettings,
+		debugID:             debugID,
 		Client:              cl,
 		Wallets:             make([]*EthereumWallet, 0),
 		headerSubscriptions: map[string]HeaderEventSubscription{},
@@ -85,7 +73,7 @@ func newEVMClient(networkSettings EVMNetwork) (EVMClient, error) {
 	if ec.NetworkConfig.Simulated {
 		ec.NonceSettings = newNonceSettings()
 	} else { // un-simulated chain means potentially running tests in parallel, need to share nonces
-		ec.NonceSettings = globalNonceManager
+		ec.NonceSettings = useGlobalNonceManager()
 	}
 
 	if err := ec.LoadWallets(networkSettings); err != nil {
@@ -111,6 +99,7 @@ func (e *EthereumClient) newHeadersLoop() {
 		if err := e.subscribeToNewHeaders(); err != nil {
 			log.Error().
 				Err(err).
+				Int("Debug ID", e.debugID).
 				Str("NetworkName", e.NetworkConfig.Name).
 				Msg("Error while subscribing to headers")
 			time.Sleep(time.Second)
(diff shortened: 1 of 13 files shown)


blockchain/optimism.go
```diff
@@ -26,154 +10,169 @@ type OptimismClient struct {
...
-		Uint64("Amount", balance.Uint64()).
-		Uint64("Estimated Gas Cost", estimatedGasCost.Uint64()).
-		Msg("Returning Funds to Default Wallet")
-	if err := o.Client.SendTransaction(context.Background(), tx); err != nil {
-		return err
-	}
-
-	return o.ProcessTransaction(tx)
-}
+// // Fund sends some OP to an address using the default wallet
```
tateexon, line 13: Do we need to keep this?

tateexon (approved):
Just one comment but otherwise it looks good to me

blockchain/ethereum.go
```diff
@@ -61,18 +58,24 @@ func newEVMClient(networkSettings EVMNetwork) (EVMClient, error) {
...
 		Wallets:             make([]*EthereumWallet, 0),
-		Nonces:              make(map[string]uint64),
 		headerSubscriptions: map[string]HeaderEventSubscription{},
 		subscriptionMutex:   &sync.Mutex{},
 		queueTransactions:   false,
 		doneChan:            make(chan struct{}),
 	}
 
+	if ec.NetworkConfig.Simulated {
+		ec.NonceSettings = newNonceSettings()
```
AnieeG, line 74: In case of load tests simulated network will also have parallel requests. Can we introduce it for simulated as well?

blockchain/optimism.go
kalverra: Not for long. Optimism Bedrock recently updated which changed a lot of their gas calculation, but they haven't fully released all the modifications yet. So I'm sitting on this just in case I need to re-use some of it. Will remove it when I get further info from Optimism team.

blockchain/ethereum.go
AnieeG: If I want to send transactions in parallel in simulated chain using same owner address, will it help in that case? I am trying to figure out if it can replace the need for -  [ConcurrentEVMClient](https://github.com/smartcontractkit/chainlink-testing-framework/blob/1e9b4179660429e489e737eae1d0c3d0efb9d0ee/blockchain/ethereum.go#L706-L709)

blockchain/nonce_settings.go
```diff
@@ -0,0 +1,144 @@
...
+		case toRegister := <-ns.registerChan:
+			if _, ok := ns.instantTransactions[toRegister.fromAddr]; !ok {
+				ns.instantTransactions[toRegister.fromAddr] = make(map[uint64]chan struct{})
+			}
+			ns.instantTransactions[toRegister.fromAddr][toRegister.nonce] = toRegister.releaseChan
+		case sentAddr := <-ns.sentChan:
+			ns.instantNoncesMu.Lock()
+			ns.instantNonces[sentAddr]++
+			ns.instantNoncesMu.Unlock()
+		case <-ns.doneChan: // Rarely need to call this
```
skudasov, line 106: It looks like a default shutdown chan, "rarely need to call this", what does it mean? Is there any more cases except shutdown of the client?

blockchain/nonce_settings.go
```diff
@@ -0,0 +1,144 @@
...
+		}
+		e.NonceSettings.Nonces[addr.Hex()] = pendingNonce
+	}
+	return e.NonceSettings.Nonces[addr.Hex()], nil
+}
+
+// watchInstantTransactions should only be called when minConfirmations for the chain is 0, generally an L2 chain.
+// This helps meter out transactions to L2 chains, so that nonces only send in order. For most (if not all) L2 chains,
+// the mempool is small or non-existent, meaning we can't send nonces out of order, otherwise the tx is instantly
+// rejected.
```
skudasov, line 91: Can we, possibly, remove the case where we can set minConfirmations = 0, I'm seeing more and more code and comments to handle and reason about that case, but I don't understand why such case as minConfirmations = 0 even exists, isn't 1 is the sane default for all networks, or even more?

blockchain/ethereum.go
```diff
@@ -28,25 +28,22 @@ import (
 	"github.com/smartcontractkit/chainlink-testing-framework/utils"
 )
 
+var debugID int
```
skudasov, line 31: It deserves some comments, what kind of `debugID` is this, or it's temporary?

blockchain/nonce_settings.go
skudasov: If we can't send `nonces` out of order, maybe we can just work in blocking mode if it's fast?

skudasov (changes requested):

blockchain/ethereum.go
kalverra: Yes, I think this can help in your case. Will make a ticket to incorporate it for that situation.

blockchain/ethereum.go
kalverra: Extraneous, removed

blockchain/nonce_settings.go
kalverra: I think this can actually be removed, as when the client shuts down, this gets a default shutdown as well. I included it assuming it would be necessary, but have found it not to be. I left it assuming that someone might need it in the future, but I can rip it out.

blockchain/nonce_settings.go
kalverra: Defaulting to `minConfirmations=1` for L2s like Optimism led to many flakes and slower test execution. It often created a race between how fast we subscribe to watch the transaction and when the block that includes the transaction is "mined." It also left us open to eternally waiting for transactions to confirm. It's possible in chains like Arbitrum to not have a block for many minutes or hours if the chain isn't active.

For using a blocking mode in all cases, I'm not fully against it, but I've no idea what the performance would look like for chains like Goerli afterwards. It probably wouldn't have much of an effect.

skudasov (approved):

PR #525, opened 2023-01-31
move config stuff to utils

skudasov:


utils/config.go
kalverra, line 1: Generic utils package

kalverra (changes requested):

utils/config.go
Tofel: I only update the name of the package since it was incorrect. In reality that change is unrelated to this PR

kalverra (approved):

PR #540, opened 2023-02-20
Add random value to mockserver

lukaszcl:


client/mockserver.go
```diff
@@ -61,6 +61,28 @@ func (em *MockserverClient) ClearExpectation(body interface{}) error {
 	return err
 }
 
+// SetRandomValuePath sets a random int value for a path
+func (em *MockserverClient) SetRandomValuePath(path string) error {
```
skudasov, line 65: Maybe it'd be of use to us but if we test feeds, please, do not use random values. You must trigger the feed every time.

skudasov (approved):

ajaskolski (approved):
lgtm

PR #565, opened 2023-03-14
Add helper util for getting an environment variable the same way in b…

tateexon:
…oth local and the remote runner without having to setup a new prefixed variable yourself. Where ever you get the variable will now set it up for the remote runner if you use this get method to do so.

### DIFF
--- a/utils/os_env.go
+++ b/utils/os_env.go
@@ -3,6 +3,8 @@ package utils
 import (
 	"fmt"
 	"os"
+
+	"github.com/smartcontractkit/chainlink-env/config"
 )
 
 // GetEnv returns the value of the environment variable named by the key
@@ -10,7 +12,7 @@ import (
 func GetEnv(key string) string {
 	val := os.Getenv(key)
 	if val != "" {
-		os.Setenv(fmt.Sprintf("TEST_%s", key), val)
+		os.Setenv(fmt.Sprintf("%s%s", config.EnvVarPrefix, key), val)
 	}
 	return val
 }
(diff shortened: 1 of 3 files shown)


utils/os_env.go
```diff
@@ -0,0 +1,16 @@
...
+import (
+	"fmt"
+	"os"
+)
+
+// GetEnv returns the value of the environment variable named by the key
+// and sets the environment variable up to be used in the remote runner
+func GetEnv(key string) string {
+	val := os.Getenv(key)
+	if val != "" {
```
kalverra, line 12: What if `val` is already prefixed with `TEST_`?

kalverra (changes requested):

utils/os_env.go
```diff
@@ -0,0 +1,16 @@
...
+		os.Setenv(fmt.Sprintf("TEST_%s", key), val)
```
tateexon, line 13: Ugh, I should reuse the config prefix instead of a string here. Fixing

utils/os_env.go
tateexon: Shouldn't matter. For example if you wanted to propagate TEST_SOMETHING to propagate it you want it to be TEST_TEST_SOMETHING. Or am I missing something else?

tateexon: I can add logic to make sure the prefixed isn't already set and send an error if so. Might prevent someone from bonking against this in a weird way in the future.

tateexon: Alright we throw an error now if there is an existing prefixed key.

kalverra (approved):

AnieeG (approved):

PR #571, opened 2023-03-16
Enables Safe EVM Header Marshaling and Celo Integration

kalverra:
**Problem 1**: Celo utilizes a weird header struct, so Geth errors when reading from it. 
**Solution**: I implemented `SafeEVMHeader` which is a minimal header struct type, and can marshal damn near any header representation. It's not Celo specific, because plenty of other chains have similar issues.

**Problem 2**: Most contract interactions call for headers to estimate gas prices, all done by Geth, and without any opportunity to use our custom, safe header value. So any contract deployments and interactions will suffer the same fate.
**Solution**: Those calls won't call for those headers if we instead make them legacy Txs by setting `GasPrice`.


contracts/ethereum/OCR2Aggregator.go
```diff
@@ -0,0 +1,3889 @@
+// Code generated - DO NOT EDIT.
```
tateexon, line 1: Can you move this to the chainlink repo? Shouldn't be adding even more contracts to the mostly deprecated charts here.

tateexon (changes requested):

tateexon (approved):

PR #608, opened 2023-05-16
don't fund the simulated client as it is already funded

AnieeG:


kalverra:
Make a comment mentioning why it's different from the standard way of funding, which might not be obvious with the concurrent client.

lukaszcl (approved):

gheorghestrimtu (approved):

PR #621, opened 2023-06-12
[TT-351] Estimate Gas

kalverra:
This PR implements a much cleaner and DRY way to incorporate our most common integration flow for a new chain. This does result in a change of the `EVMClient` interface, adding the below methods to it.

```go
NewTx(
  fromPrivateKey *ecdsa.PrivateKey,
  nonce uint64,
  to common.Address,
  value *big.Int,
  gasEstimations GasEstimations,
) (*types.Transaction, error)
EstimateGas(callMsg ethereum.CallMsg) (GasEstimations, error)
```

These enable us to easily write new gas estimation methods, along with switching between Dynamic and Legacy transaction types, a common integration request.


blockchain/ethereum.go
```diff
@@ -85,6 +85,22 @@ func newEVMClient(networkSettings EVMNetwork) (EVMClient, error) {
...
 	go ec.newHeadersLoop()
 
+	// Check if the chain supports EIP-1559
+	// https://eips.ethereum.org/EIPS/eip-1559
+	if _, err = cl.SuggestGasTipCap(context.Background()); err != nil {
+		if strings.Contains(err.Error(), "unknown field") ||
+			strings.Contains(err.Error(), "does not exist") ||
+			strings.Contains(err.Error(), "is not available") {
+			ec.supportsEIP1559 = false
+			log.Debug().Msg("Chain does not support EIP-1559 transactions, using legacy")
```
AnieeG, line 95: Can we not simply take it as a network input?

blockchain/klaytn.go
```diff
@@ -26,142 +16,154 @@ type KlaytnClient struct {
...
-		Str("Token", "KLAY").
-		Str("From", fromAddress.Hex()).
-		Str("Amount", balance.String()).
-		Msg("Returning Funds to Default Wallet")
-	if err := k.SendTransaction(context.Background(), tx); err != nil {
-		return err
-	}
-	return k.ProcessTransaction(tx)
-}
+// // Fund overrides ethereum's fund to account for Klaytn's gas specifications
```
AnieeG, line 31: Do you want to keep these?

blockchain/known_networks.go
```diff
@@ -35,7 +35,7 @@ func wrapSingleClient(networkSettings EVMNetwork, client *EthereumClient) EVMCli
 	case ArbitrumClientImplementation:
 		wrappedEc = &ArbitrumClient{client}
 	case OptimismClientImplementation:
-		wrappedEc = &OptimismClient{client}
+		wrappedEc = &OptimismClient{*client}
```
AnieeG, line 38: Why is Optimism different?

blockchain/ethereum.go
kalverra: We can go that way too, though I think we'd need to do

1. Add another config field
2. Add another method to the EVMClient interface
3. Deal with possible cases of interface shadowing, which can be frustrating

AnieeG: The network config seems cleaner and robust rather than calling the function and trying to parse the error message.

AnieeG (approved):

PR #629, opened 2023-06-20
[TT-373] Improves Rpc Reliability

kalverra:
Fortifies our main test subscription loop with some better resubscription logic. This includes some timeouts when asking for data from the chain, which often led to hanging requests in the old style.

This also adds some connection status indicators to the `EVMClient` interface that tests can pull from to determine what to do when the connection is dropped, and can resume business as usual otherwise.

```go
// Connection Status
// ConnectionIssue returns a channel that will receive a timestamp when the connection is lost
ConnectionIssue() chan<- time.Time
// ConnectionRestored returns a channel that will receive a timestamp when the connection is restored
ConnectionRestored() chan<- time.Time
```

We also reduce our calls to gas estimation. We used to ask for a gas estimation every block, which could potentially lead to rate limits, especially on fast-moving L2s. The new general philosophy with these changes is to treat active chain monitoring (like subscribing to events) as a secondary way of checking the chain. Especially for long-running tests, unless absolutely necessary, you should wait until the end of the test to check on-chain state in order to avoid RPC flakes.


AnieeG (approved):

blockchain/transaction_confirmers.go
```diff
@@ -301,73 +302,158 @@ func (e *EthereumClient) GetHeaderSubscriptions() map[string]HeaderEventSubscrip
...
+		}
+	}
+}
+
+// resubscribeLoop polls the RPC connection until it comes back up, and then resubscribes to new headers
+func (e *EthereumClient) resubscribeLoop(headerChannel chan *SafeEVMHeader, lastHeaderNumber uint64) ethereum.Subscription {
+	rpcDegradedTime := time.Now()
+	pollInterval := time.Millisecond * 500
+	checkTicker, informTicker := time.NewTicker(pollInterval), time.NewTicker(time.Second*10)
+	reconnectAttempts, consecutiveSuccessCount, targetSuccessCount := 0, 0, 3
```
AnieeG, line 348: Nit : All magic numbers in constants would help

kalverra: I can make the target count a CONST, unfortunately the other two must stay as it.

PR #631, opened 2023-06-20
Add estimated finality time calculations

AnieeG:


blockchain/ethereum.go
```diff
@@ -820,6 +822,55 @@ func (e *EthereumClient) GetLatestFinalizedBlockHeader(ctx context.Context) (*ty
...
+	log.Info().Msg("Calculating estimated finalization time")
+	if e.NetworkConfig.FinalityTag {
+		currentFinalizedHeader, err := e.GetLatestFinalizedBlockHeader(ctx)
+		if err != nil {
+			return 0, err
+		}
+		now := time.Now()
+		// wait for the next finalized block
+		for {
+			time.Sleep(30 * time.Second)
```
kalverra, line 838: Seems like you could potentially miss a lot of other finalized headers here. Is it possible to do a historic lookback, instead of just waiting for new blocks as we run the test?

AnieeG: yeah. There is a slight chance that I miss a finalized block if it's a super fast chain. I tried to look for last to last finalized block or have a subscriber which can do something like newFinalizedHead (just like newHead) but didn't find any :(

blockchain/ethereum.go
```diff
@@ -820,6 +822,60 @@ func (e *EthereumClient) GetLatestFinalizedBlockHeader(ctx context.Context) (*ty
...
+				return 0, errors.New("timed out waiting for next finalized block")
+			}
+		}
+	} else {
+		blckTime, err := e.AvgBlockTime(ctx)
+		if err != nil {
+			return 0, err
+		}
+		if e.NetworkConfig.FinalityDepth == 0 {
+			return 0, errors.New("finality depth is 0 and finality tag is not enabled")
```
kalverra, line 867: Just want a little more detail on this error: what should I do to fix it?

blockchain/ethereum.go
```diff
@@ -820,6 +822,60 @@ func (e *EthereumClient) GetLatestFinalizedBlockHeader(ctx context.Context) (*ty
...
+	if e.NetworkConfig.FinalityTag {
+		currentFinalizedHeader, err := e.GetLatestFinalizedBlockHeader(ctx)
+		if err != nil {
+			return 0, err
+		}
+		c, cancel := context.WithTimeout(ctx, MaxTimeoutForFinality)
+		defer cancel()
+		tick := time.NewTicker(time.Second)
+		// wait for the next finalized block
+		for {
```
kalverra, line 842: Maybe worth a separate method to organize?

kalverra (changes requested):

kalverra (approved):

PR #643, opened 2023-07-20
LogWatch, stream and notify about docker logs in tests

skudasov:
LogWatch is a small component to:
- stream logs from your local docker setup to Loki
  example format:
  ```
  {type="log_watch", container="container-0",test="TestExampleLokiStreaming/nobody_expects_the_spammish_repetition"}
  ```
- use regex to match some log lines and stop the test

Check the tests for more examples how to use


utils/common.go
```diff
@@ -21,3 +21,15 @@ func GetTestLogger(t *testing.T) zerolog.Logger {
 	l := zerolog.New(zerolog.NewTestWriter(t)).Output(zerolog.ConsoleWriter{Out: os.Stderr}).Level(lvl).With().Timestamp().Logger()
 	return l
 }
+
+// GetTestComponentLogger instantiates a logger that takes into account the test context, log level and a component name
+func GetTestComponentLogger(t *testing.T, name string) zerolog.Logger {
```
kalverra, line 26: Just wondering how this differs from our current `GetLogger`?

logwatch/logwatch.go
```diff
@@ -0,0 +1,181 @@
...
+		Msg("Received notification from container")
+	return msg
+}
+
+// OnMatch calling your testing hook on first match
+func (m *LogWatch) OnMatch(f func()) {
+	go func() {
+		for {
+			msg := <-m.notifyTest
+			m.log.Warn().
```
kalverra, line 67: Why is this a warn log?

logwatch/logwatch_user_loki_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+/* These tests are for testing Loki format, they rarely change so you can run them manually */
+
+func TestExampleLokiStreaming(t *testing.T) {
+	t.Skip("uncomment and run manually")
+	t.Run("stream all container logs to Loki, subtest 1", func(t *testing.T) {
+		testData := testData{repeat: 1, perSecond: 0.01, streams: []string{"A\nB\nC\nD", "E\nF\nG\nH"}}
+		d, err := NewDeployment(testData)
+		// nolint
+		defer d.Shutdown()
```
kalverra, line 20: You're risking nil pointer errors here. Check error before calling shutdown.

kalverra (changes requested):

logwatch/logwatch.go
```diff
@@ -0,0 +1,181 @@
...
+	loki       *wasp.LokiClient
+	patterns   map[string][]*regexp.Regexp
+	notifyTest chan *LogNotification
+	containers []testcontainers.Container
+	consumers  map[string]*ContainerLogConsumer
+}
+
+// NewLogWatch creates a new LogWatch instance, with a Loki client
+func NewLogWatch(t *testing.T, patterns map[string][]*regexp.Regexp) (*LogWatch, error) {
+	loki, err := wasp.NewLokiClient(wasp.NewEnvLokiConfig())
```
tateexon, line 38: Would it make sense to move the loki client into the chainlink-env code base since it is part of our env stuff? Would also mean we don't need to pull in wasp for everything that uses chainlink-env.

skudasov: We can move it later since we remove chainlink-env in the future.
Also, I'm using it from `wasp` just because of it's `vendored`. Just to not cause more dep hell.

utils/common.go
skudasov: I can do it with `functional options` and have one function. Basically, it's adding a `component` field. But I'm not sure you need it.

logwatch/logwatch_user_loki_test.go
skudasov: fixed

tateexon (approved):

logwatch/logwatch.go
skudasov: Usually, we'll search for specific or common panics/fatals, etc. Since it's not our `error`, `warn` seems ok.

utils/common.go
skudasov: Okay, can skip this function and just do
```
	l := utils.GetTestLogger(t).With().Str("Component", "LogWatch").Logger()
```

AnieeG (approved):

tateexon (approved):

kalverra (approved):

PR #649, opened 2023-07-27
[TT-479] - adding finality validations

AnieeG:


blockchain/ethereum.go
```diff
@@ -536,6 +537,69 @@ func (e *EthereumClient) ProcessEvent(name string, event *types.Log, confirmedCh
...
+	}
+	txHdr, err := e.HeaderByNumber(context.Background(), receipt.BlockNumber)
+	if err != nil {
+		return nil, time.Time{}, err
+	}
+	finalizer := NewTransactionFinalizer(e, txHdr, receipt.TxHash)
+	key := "txFinalizer-" + txHash.String()
+	e.AddHeaderEventSubscription(key, finalizer)
+	defer e.DeleteHeaderEventSubscription(key)
+	err = finalizer.Wait()
```
kalverra, line 572: Are there situations where you want to wait for multiple finalized txs and then call the general `WaitForEvents`?

blockchain/ethereum.go
```diff
@@ -536,6 +537,69 @@ func (e *EthereumClient) ProcessEvent(name string, event *types.Log, confirmedCh
...
+		return nil, time.Time{}, err
+	}
+	return finalizer.FinalizedBy, finalizer.FinalizedAt, nil
+}
+
+// IsTxFinalized checks if the transaction is finalized on chain
+// if the tx is not finalized it returns false, the latest finalized header number and the time at which it was finalized
+// if the tx is finalized it returns true, the finalized header number by which the tx was considered finalized and the time at which it was finalized
+// In case of simulated network, it always returns true
+func (e *EthereumClient) IsTxFinalized(txHdr, header *SafeEVMHeader) (bool, *big.Int, time.Time, error) {
```
kalverra, line 583: Naming: you're only comparing headers here, I think it's more apt to call this `IsHeaderFinalized`, and maybe also add `IsTxFinalized` that takes in just a tx hash and checks if that would be useful?

blockchain/ethereum.go
AnieeG: Including multiple transactions for finality check can lead to unpredictable delays. That's why not including it in `WaitForEvents` was a conscious decision. It's clean to do it by individual transactions. You generally wouldn't want to to wait for finality for generic transactions.

blockchain/ethereum.go
AnieeG: To know if a Tx is finalized you will need the block number from its receipt. unless a tx is confirmed it can't be finalized.  From that point `IsTxFinalized` will always mean `IsTxHeaderFinalized`. So we won't need `IsTxFinalized` but I can change the naming to reflect `IsHeaderFinalized`

kalverra (approved):

PR #687, opened 2023-09-07
Keep trying for geth container start

AnieeG:


tateexon (approved):

docker/docker.go
```diff
@@ -28,3 +30,21 @@ func CreateNetwork() (*tc.DockerNetwork, error) {
 	log.Trace().Any("network", dockerNetwork).Msgf("created network")
 	return dockerNetwork, nil
 }
+
+func StartContainerWithRetry(req tc.ContainerRequest) (tc.Container, error) {
```
tateexon, line 34: Would it make more sense to pass in a tc.GenericContainerRequest instead. It gives us more flexibility if in the future we need that part to be different for a specific container.

tateexon (approved):

PR #693, opened 2023-09-11
increase host cfg

AnieeG:


skudasov (approved):

docker/docker.go
```diff
@@ -45,6 +46,12 @@ func StartContainerWithRetry(req tc.GenericContainerRequest) (tc.Container, erro
 			log.Error().Err(err).Msgf("Cannot terminate %s container to initiate restart", req.Name)
 			return nil, err
 		}
+		// spike the host config resources
+		req.HostConfigModifier = func(config *container.HostConfig) {
```
tateexon, line 50: I think we should do this differently since this method can be used for more than just geth. We should pass in the wanted host config modifier instead. I need to use this retry for chainlink nodes and I do not need anything close to this big of a cpu count.

PR #714, opened 2023-09-28
Added ZkSync methods

smickovskid:
- Added ZkSync methods


friedemannf:
I added an override for the funding/refunding logic for zkSync. I wasn't sure whether to opt for duplicating code or changing the existing implementation on `EthereumClient` but chose the former for now.
Maybe it would make more sense for 
```
func attemptReturn(e *EthereumClient, fromKey *ecdsa.PrivateKey, attemptCount int) (*types.Transaction, error)
```
to accept an interface? Thereby we could prevent having the same code twice.

Tofel:
closed due to staleness; we have also decided to handle ZKSync deployments differently

PR #725, opened 2023-10-10
Add option to set custom postgres image version

lukaszcl:


docker/test_env/postgres.go
```diff
@@ -36,17 +37,26 @@ func WithPostgresDbContainerName(name string) PostgresDbOption {
...
-		User:     "postgres",
-		Password: "mysecretpassword",
-		DbName:   "testdb",
-		Port:     "5432",
-		l:        log.Logger,
+		User:         "postgres",
+		Password:     "mysecretpassword",
+		DbName:       "testdb",
+		Port:         "5432",
+		ImageVersion: "15.3",
```
AnieeG, line 58: Let's make it a constant

AnieeG (approved):

docker/test_env/postgres.go
lukaszcl: hm, will it help with anything? Btw, if you think this value should be const, other should be as well right?

AnieeG: Yeah other should be as well. Just a good to have stuff. Feel free to ignore now. can be a follow up PR later. It just makes it easy to update and easily readable.

PR #741, opened 2023-10-19
[WIP] [TT-590] add support for multiple log targets to LogWatch

Tofel:
This PR introduces the ability to easy add new logging targets while adding two of them: Loki and local files.

It is possible to use any number of logging targets concurrently. By default no logging targets, they need to be specified using env variable `LOGWATCH_LOG_TARGETS`, which currently accepts two values: `file` and `loki` (case insensitive).

LogWatch's API has not changed, but now the constructor accepts variadic options that can be used to override default implementations for log target handlers (useful for example for unit testing if we don't want to create any files or send anything to Loki but still make sure that correct handler was called expected number of times).

The constructor also executes some log target validations and operations:
* is there a known handler for every requested log target?
* deactivate known handlers that weren't requested/set via environment variables

Also, Loki client is now lazy initalised the first time it is requested instead of being always initialised in the constructor (even if streaming logs to Loki was not requested).

Last, but not least... file logs are also streamed instead of being saved at the end of test, which means that in case of container failure we will still have some logs which might help to debug failure cause.

When a test ends each active handler prints information about log location.

It is recommended to connect/disconnect LogWatch to containers using PostStarts/PostStops hooks instead of doing it manually.

**Known issues existing in previous versions:**
* slow containers might result in `panic` when log producer cannot connect to a container within 5 seconds (context timeout hardcoded in `testcontainers-go`, there's a [PR open](https://github.com/testcontainers/testcontainers-go/pull/1278) that adds retries (although it seems from my experiments that if log producer fails to start in 5 seconds there's something off with the container and retrying won't help anyway)
* if the test is too short logs might not be sent to Loki before test finishes (we should consider adding a way to flush them on shutdown)

**Breaking changes:**
* env variable `TEST_LOG_COLLECT ` is no longer used. To indicate that you want to collect logs simply add targets to `LOGWATCH_LOG_TARGETS` or leave it empty to skip log collection

**New env variables:**
* `GRAFANA_URL` -- URL to Grafana instance that is connected to Loki instance to which logs where streamed
* `GRAFANA_DATASOURCE` -- Loki datasource id that will be used when building the URL

**What's missing:**
* tests for new functionality
* validating whether Grafana-related env vars were set
* creating/reading a global execution id that will be send together with logs and allow us to filter them in Grafana
* building of Grafana Dashboard URL instead of Explore URL (as is currently implemented)

Example of changes to `chainlink` repo that uses `testcontainers-go` hooks to start and stop listening to logs can be found [here](https://github.com/smartcontractkit/chainlink/pull/11007).


logwatch/logwatch.go
```diff
@@ -25,30 +27,101 @@ type LogNotification struct {
...
-		loki:       loki,
-		patterns:   patterns,
-		notifyTest: make(chan *LogNotification, 10000),
-		consumers:  make(map[string]*ContainerLogConsumer, 0),
-	}, nil
+
+	return logWatch, nil
+}
+
+func (m *LogWatch) validateLogTargets() error {
```
lukaszcl, line 78: @Tofel maybe it's better to make it a function (not method) that takes log targets? We could call it before creating `LogWatch` object in 58th line

logwatch/logwatch_handlers.go
```diff
@@ -0,0 +1,173 @@
...
+}
+
+// streams logs to local files
+type FileLogHandler struct {
+	testLogFolders    map[string]string
+	shouldSkipLogging map[string]bool
+}
+
+func (h FileLogHandler) Handle(c *ContainerLogConsumer, content LogContent) error {
+	if val, ok := h.shouldSkipLogging[content.TestName]; val && ok {
```
lukaszcl, line 47: @Tofel maybe we could get rid of `shouldSkipLogging` field by removing the handler from `logTargetHandlers` in logwatch.go when there was an error? I think this would simplify the code in this file. WDYT?

Tofel: it would indeed! I tried to be smart here, so that if folder creation for one test fails the rest will try to create their folders anyway, but what are the chances of that actually happening? I guess very low.

logwatch/logwatch.go
Tofel: I don't see why not 👍

logwatch/logwatch_handlers.go
```diff
@@ -0,0 +1,173 @@
...
+
+	grafanaUrl := GrafanaExploreUrl{
+		baseurl:    os.Getenv("GRAFANA_URL"),
+		datasource: os.Getenv("GRAFANA_DATASOURCE"),
+		queries:    queries,
+		rangeFrom:  rangeFrom.UnixMilli(),
+		rangeTo:    rangeTo.UnixMilli() + 60000, //just to make sure we get the last message
+	}.getUrl()
+
+	m.log.Info().Str("URL", string(grafanaUrl)).Msg("Loki logs can be found in Grafana at (will only work when you unescape quotes):")
```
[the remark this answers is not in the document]
lukaszcl: @Tofel btw, wdyt about having a function to get the grafana url? This way the clients can show it however they want (for example in Github CI run summary). It's also not depended on the `m.log`

logwatch/logwatch_handlers.go
lukaszcl: yeah, probably very low. Also, have you considered having separate `FileLogHandler` for each test?

lukaszcl: > EDIT: on the other hang with streaming logs to Loki it's probable that connection is intermitently lost and then re-established. If we remove the handler we will lose all logs since the time of first failure. If we keep a map per test we might still get logs from other tests.

Yeah, that makes sense!

logwatch/logwatch.go
lukaszcl: that's ok. We can leave it then!

logwatch/logwatch_handlers.go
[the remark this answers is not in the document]
Tofel: I think it's a very good idea, let me look into it

PR #742, opened 2023-10-19
update go-ethereum to one without known vulnerability, ignore one coming from wasp

Tofel:
why ignore?

the library with vulnerablilty is used by our Loki dependency, which is in version `v1.6.2`, but latest release is `v.2.9.2` so I didn't want to get into updating the major version (yet)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,7 @@ require (
 	github.com/avast/retry-go v3.0.0+incompatible
 	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc
 	github.com/docker/go-connections v0.4.0
-	github.com/ethereum/go-ethereum v1.12.2
+	github.com/ethereum/go-ethereum v1.12.0
 	github.com/go-resty/resty/v2 v2.7.0
 	github.com/google/uuid v1.3.1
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -6,7 +6,7 @@ require (
 	github.com/avast/retry-go v3.0.0+incompatible
 	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc
 	github.com/docker/go-connections v0.4.0
-	github.com/ethereum/go-ethereum v1.12.0
+	github.com/ethereum/go-ethereum v1.12.2
```
tateexon, line 9: I thought we had issues with this change due to the chainlink repo not having updated yet. Please very this change doesn't break e2e tests in core before we merge this.

Tofel: so much for them using semver?

Tofel: @tateexon good call, this field doesn't exist in `toml/config.go` of `go-ethereum` anymore:
```
txpool.DefaultConfig
```

I'll add that `geth` vulnerability to ignored ones and create a ticket in JIRA to remove it once `chainlink-core` [has bumped the dependency](https://github.com/smartcontractkit/chainlink/pull/10264). I don't like having vulnerability check always red.

PR #743, opened 2023-10-20
[TT-561] move chainlink-env as a package

Tofel:


AnieeG:
is there a way to break this up in multiple PRs?

Tofel:
> is there a way to break this up in multiple PRs?

I'd rather not, although I know it's huge. Would it help if I went through commits and listed all files that were modified instead of copied?

.github/workflows/env-publish-test-bash-image.yaml
```diff
@@ -0,0 +1,27 @@
+name: Publish Test Base Image (Env)
```
tateexon, line 1: The filename has a typo, should be ...test-base-image.... not bash

env/logging/log.go
```diff
@@ -0,0 +1,40 @@
+package logging
```
tateexon, line 1: This file can go away in favor of the logging package already in CTF: https://github.com/smartcontractkit/chainlink-testing-framework/blob/main/logging/log.go
No need to have two separate ones

Tofel: good catch!

Tofel:
Moved to https://github.com/smartcontractkit/chainlink-testing-framework/pull/747 where changes are better visible

PR #747, opened 2023-10-24
[TT-561] move chainlink-env as a package (commit-splitted)

Tofel:
Changes:
* `chainlink-env` code needs migrated and converted to a package ✅
* all import paths updated ✅
* fixed all issues reported by linterns ✅
** `imports` and `examples` excluded from analysia
* the ci and tests migrated and passing ✅
* Dockerfiles adjusted and working ✅
* Infra permissions will need to be added to chainlink-testing-framework to push the expected docker images into ECR ❓ @tateexon 
* `chainlink` tests work using migrated code: [PR](https://github.com/smartcontractkit/chainlink/pull/11038)


tateexon:
It might make sense to put the chainlink-env package into a folder called k8s instead of env. My reasoning for this is that we have a folder called docker which hosts the docker specific stuff so it makes sense if you want to look for k8s specific stuff you would open the k8s folder instead of env.

env/.github/workflows/e2e.yaml
```diff
@@ -0,0 +1,118 @@
+name: Deployments tests (E2E)
```
tateexon, line 1: We shouldn't need the .github directory and these files within since the ones needed should be moved to the .github at the base of the chainlink-testing-framework repo.

Tofel: argh you! merge issue...

.github/workflows/env-publish-test-base-image.yaml
```diff
@@ -0,0 +1,27 @@
...
+on:
+  push:
+    tags:
+      - 'v*'
+    paths:
+    - "k8s/**"
+
+jobs:
+  publish_test_base_image:
+    if: ${{ false }}
```
tateexon, line 11: Is this still needed?

Tofel: nope! and that reminds that I have to restore old triggers for `chainlink-env` specific jobs

.github/workflows/env-publish-test-base-image.yaml
```diff
@@ -0,0 +1,27 @@
+name: (k8s package) Publish Test Base Image
+on:
+  push:
+    tags:
+      - 'v*'
+    paths:
```
tateexon, line 6: This won't work as expected since new tags/releases can have more than 1 commit associated with them. The way you currently have the logic this would only work if the k8s changes happened to happen in the commit that the tag is based on. If the k8s changes happened in the commit just before it then it would not create a new release.

Tofel: do you think we need some kind of more sophisticated filter here or we should just run it when when a tag is added?

tateexon (commented):
The part I am unsure about is whether we should add this as a package or just migrate the code in so we only have the 1 go project. They both have their pros and cons. 1 go project is easier to work within but the separation of the two may make it easier for projects that only want to import 1 part or the other. I think most projects will use both anyways so it may make sense to have it just be 1 project instead of two. @AnieeG @skudasov @kalverra @gheorghestrimtu Thoughts?

Tofel:
@tateexon also thought about that and checked how many projects use only `chainlink-env` and yes, there were a couple (e.g. Atlas from @gheorghestrimtu).

In the end I decided against having `env` as a separate module, because:
* it would have to share the same tag as CTF unless we went with git subtree, which would increase complexity (e.g. we couldn't commit changes to CTF and k8s in the same commit)
* CTF isn't that heavy of a module/repo to worry about having to download it and it's dependency, when you only need k8s code
and a couple of others that I cannot recall now (had them written down, but didn't save/share).

Anyway let me know your thoughts guys.

Also, I was thinking that if we stick to packages `qa-charts` could be put in `k8s/charts`?

lukaszcl:
@tateexon @Tofel  +1 for a single go project. I think it makes more sense here. Also, I think we should have one CTF tag

tateexon (dismissed):

tateexon (approved):

docs/setup/code.md
```diff
@@ -7,11 +7,11 @@ parent: Setup
 
 # Test Setup Code
 
-Now that we've got our config and Kubernetes sorted, we can write a bit of code that will deploy an environment for our test to run. To deploy our simulated geth, mock-server, and Chainlink instances, we rely on another chainlink library, [chainlink-env](https://github.com/smartcontractkit/chainlink-env/). This library handles deploying everything our test needs to the Kubernetes cluster.
+Now that we've got our config and Kubernetes sorted, we can write a bit of code that will deploy an environment for our test to run. To deploy our simulated geth, mock-server, and Chainlink instances, we rely on `env` package. This package handles deploying everything our test needs to the Kubernetes cluster.
 
 ```go
-// We use the chainlink-env library to make and handle deployed resources
-import "github.com/smartcontractkit/chainlink-env/environment"
+// We use the env package to make and handle deployed resources
```
AnieeG, line 13: The package name in comment

AnieeG (approved):

lukaszcl (approved):

sebawo:
> @tateexon @Tofel +1 for a single go project. I think it makes more sense here. Also, I think we should have one CTF tag.
> 
> >

I agree we should have one release cycle for CTF+chainlink-env+qa-charts. This is one of the motivations to merge repos and have separate releases for modules.

PR #749, opened 2023-10-27
Add tests for MustGetSelectedNetworksFromEnv and NewEVMNetwork

lukaszcl:
- [x] add tests for NewEVMNetwork

### DIFF
--- a/networks/known_networks_test.go
+++ b/networks/known_networks_test.go
@@ -17,21 +17,17 @@ func TestMain(m *testing.M) {
 }
 
 func TestMustGetSelectedNetworksFromEnv_Missing_SELECTED_NETWORKS(t *testing.T) {
-	defer os.Clearenv()
-
 	require.Panics(t, func() {
 		MustGetSelectedNetworksFromEnv()
 	})
 }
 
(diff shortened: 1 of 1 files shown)


networks/known_networks_test.go
```diff
@@ -0,0 +1,85 @@
...
+		MustGetSelectedNetworksFromEnv()
+	})
+}
+
+func TestMustGetSelectedNetworksFromEnv_Missing_HTTP_URLS(t *testing.T) {
+	defer os.Clearenv()
+
+	networkKey := "ARBITRUM_GOERLI"
+
+	os.Setenv("SELECTED_NETWORKS", networkKey)
```
tateexon, line 32: https://pkg.go.dev/testing#T.Setenv
Using t.Setenv will set the environment variable within the scope of the test and the test will clear it automatically without the need to defer the cleanup.

tateexon (changes requested):

networks/known_networks_test.go
lukaszcl: great idea! Pushed a new commit

tateexon (approved):

PR #754, opened 2023-10-31
[TT-675 TT-676]Fix Test/Prefix Defaults and Fix Base Image Node Download 

tateexon:


k8s/presets/presets.go
```diff
@@ -10,9 +10,38 @@ import (
 	"github.com/smartcontractkit/chainlink-testing-framework/k8s/pkg/helm/reorg"
 )
 
+var BaseToml = `[Log]
```
kalverra, line 13: Are we able to use the config-builder logic we use for docker envs and just `.String()` it instead?

tateexon: Not unless we want to import chainlink into this repository, which we don't.

kalverra (approved):

PR #755, opened 2023-11-01
Lazy Fund Return

kalverra:
Cause it's testnet and I'm tired of dealing with it.


tateexon (approved):

utils/os_env.go
```diff
@@ -13,8 +13,8 @@ func GetEnv(key string) (string, error) {
 	val := os.Getenv(key)
 	if val != "" {
 		prefixedKey := fmt.Sprintf("%s%s", config.EnvVarPrefix, key)
-		if os.Getenv(prefixedKey) != "" {
-			return val, fmt.Errorf("environment variable collision with prefixed key, Original: %s, Prefixed: %s", key, prefixedKey)
+		if os.Getenv(prefixedKey) != "" && os.Getenv(prefixedKey) != val {
+			return val, fmt.Errorf("environment variable collision with prefixed key, Original: %s=%s, Prefixed: %s=%s", key, val, prefixedKey, os.Getenv(prefixedKey))
```
AnieeG, line 17: Does this mean evm_key will always have to match with <network>_key?

kalverra: Not quite. Before behavior was, if you had `X_VAR` set, you would get an error if you also had `TEST_X_VAR` set, no matter what they were, as conflicts could lead to confusing behavior.

Now, if you have `X_VAR` and `TEST_X_VAR` set to the same value, we'll allow it.

AnieeG (approved):

PR #756, opened 2023-11-02
Post chainlink-env/qa-charts merge cleanup

tateexon:
- Use generic Ptr function in place of alias Str and Num 
- Add ability to use local charts
- Add test that uses local charts
- Remove non existent paths from utils/project_path.go
- Remove polygon-edge since it isn't used
- Add pre-commit hook installation to makefile and add golangci-lint to hooks


k8s/e2e/common/test_common.go
```diff
@@ -179,8 +179,12 @@ func Test5NodesSoakEnvironmentWithPVCs(t *testing.T) {
 	})
 }
 
-func TestWithSingleNodeEnv(t *testing.T) {
+func TestWithSingleNodeEnvParallel(t *testing.T) {
```
kalverra, line 182: I'm just curious what's the point of having a separate one for this?

kalverra (approved):

k8s/e2e/common/test_common.go
tateexon: I need to set an environment variable in the ones that tests with local qa-charts. You can't set environment variables in parallel tests since they will affect other tests running in parallel.

PR #764, opened 2023-11-07
TT-678: Eth2: Geth + Prysm in Docker and k8s

Tofel:
Description:
* it uses latest versions of `Geth` and `Prysm` to start `eth2` network
* it still supports launching old `geth` PoW network 
* it provides a builder for Ethereum network creation using Docker (with the idea to later extend it to support other execution layer clients like Besu or Erigon, other consensus layer solutions like Lighthouse)
* for now only 1 node in each layer is created
* some beacon chain config values can be overwritten (e.g. `SecondsPerSlot` to make the chain faster)

Known issues:
* impossible to start Prysm beacon chain with `--interop-eth1data-votesgeth` flag... puzzling, to say the least, because it works, when used in `docker-compose.yml` and it also works, when executed from inside container (when started with `tail -f /dev/null` entrypoint), having said that it doesn't seem to impact anything
* unresolved error for `gp2` and `gp3` storage classes in `sdlc` cluster: 
```
running PreBind plugin "VolumeBinding": Operation cannot be fulfilled on persistentvolumes "eth2-0-1699628290-chain-state-storage": the object has been modified; please apply your changes to the latest version and try again
```

Tentative support:
* using an existing environment config instead of creating it from a scratch (the code is there, but I haven't tested it out yet, because in `chainlink` repo there are no functions to save an existing config and I didn't have time to add one yet)

In the future:
* remove `PrivateChain []test_env.PrivateChain // for tests using non-dev networks` from builders in other repos and have a unified approach to building simulated chains (based on the builder I introduced here); in the short-time keep the API but make it use new builder underneath

Still missing:
* support for restarting the environment after stopping it (if we even need that, as that would require to remove reaper so it doesn't delete stopped containers)

Sample PR using it: https://github.com/smartcontractkit/chainlink/pull/11214


docker/test_env/ethereum_env.go
```diff
@@ -0,0 +1,442 @@
...
+	existingConfig    *EthereumNetworkConfig
+}
+
+type Eth2Components struct {
+	Geth        *Geth2
+	BeaconChain *PrysmBeaconChain
+	Validator   *PrysmValidator
+}
+
+func NewEthereumNetworkBuilder(t *testing.T) EthereumNetworkBuilder {
```
lukaszcl, line 58: @Tofel can we decouple it from `testing.T`? This way we can use it outside of `go test`. For example in a CLI to start test env

docker/test_env/ethereum_env.go
```diff
@@ -0,0 +1,442 @@
...
+	b.dockerNetworks = networks
+	return b
+}
+
+func (b *EthereumNetworkBuilder) WithExistingConfig(config EthereumNetworkConfig) *EthereumNetworkBuilder {
+	b.existingConfig = &config
+	return b
+}
+
+func (b *EthereumNetworkBuilder) Build() error {
```
lukaszcl, line 108: @Tofel `Build()` usually in the [builder pattern](https://refactoring.guru/design-patterns/builder/go/example) usually returns the object that was built. The builder itself is only used to simplify initialisation of the final object. Here I'd expect to use `EthereumNetworkBuilder` to build `EthereumNetwork` instance

docker/test_env/ethereum_env.go
```diff
@@ -0,0 +1,442 @@
...
+	b.executionNodes = executionNodes
+	return b
+}
+
+func (b *EthereumNetworkBuilder) WithBeaconChainConfig(config BeaconChainConfig) *EthereumNetworkBuilder {
+	b.beaconChainConfig = &config
+	return b
+}
+
+func (b *EthereumNetworkBuilder) UsingDockerNetworks(networks []string) *EthereumNetworkBuilder {
```
lukaszcl, line 98: @Tofel any reason to use `Using` instead of `With` here? Just trying to understand the difference :)

docker/test_env/geth2.go
```diff
@@ -0,0 +1,302 @@
...
+	ExternalHttpUrl      string
+	InternalHttpUrl      string
+	ExternalWsUrl        string
+	InternalWsUrl        string
+	InternalExecutionURL string
+	ExternalExecutionURL string
+	ExecutionDir         string
+	consensusLayer       ConsensusLayer
+	l                    zerolog.Logger
+	t                    *testing.T
```
lukaszcl, line 46: @Tofel can we decouple testing.T from this docker wrapper? see https://github.com/smartcontractkit/chainlink-testing-framework/pull/764/files#r1389077799

charts/geth-prysm/templates/geth.configmap.yaml
```diff
@@ -0,0 +1,122 @@
...
+        "terminalTotalDifficultyPassed": true
+      },
+      "nonce": "0x0",
+      "timestamp": "0x6548d44d",
+      "extraData": "0x0000000000000000000000000000000000000000000000000000000000000000123463a4b065722e99115d6c222f267d9cabb5240000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
+      "gasLimit": "0x1c9c380",
+      "difficulty": "0x1",
+      "mixHash": "0x0000000000000000000000000000000000000000000000000000000000000000",
+      "coinbase": "0x0000000000000000000000000000000000000000",
+      "alloc": {
```
AnieeG, line 38: might need more accounts here. If we can add accounts with balance later that will work as well. 
Use Case - There are tests where we use close to 20 networks in one test. In that case for parallel contract deployments across those multiple networks if we use similar owner account the contract address becomes same for multiple networks. CL node cannot support same contract address for different networks yet. We might need around 20 accounts here so that the test can use diff owner accounts for diff networks.

docker/test_env/ethereum_env.go
Tofel: we could, but then logger won't be associated to any test, but what impact would that have?

docker/test_env/ethereum_env.go
Tofel: I know, but didn't want to overcomplicate it... oh well, let me change it as I had to create `EhtereumNetworkConfig` anyway

charts/geth-prysm/templates/geth.configmap.yaml
Tofel: yes, I can do that the same way it's done for besu :-)

docker/test_env/ethereum_env.go
lukaszcl: @Tofel hm, if `testing.T` is only needed for a logger then maybe have `WithLogger()` that takes a logger that already has `testing.T`?

Tofel: okay, I found a way that should work

skudasov (approved):

PR #771, opened 2023-11-17
A cpuple of eth2 updated

Tofel:
* makde validator count configurable in eth2
* updated go-ethereum image to the latest


Tofel:
Included also in https://github.com/smartcontractkit/chainlink-testing-framework/pull/782

PR #780, opened 2023-11-22
Use killgrave 0.4.1

lukaszcl:
New version of killgrave was recently released. But, we are not compatible with it ([errors](https://github.com/smartcontractkit/chainlink/actions/runs/6948551353/job/18912175297?pr=11355)). This PR hardcodes the old version that we used in the past.

### DIFF
--- a/docker/test_env/killgrave.go
+++ b/docker/test_env/killgrave.go
@@ -136,7 +136,7 @@ func (k *Killgrave) StartContainer() error {
 }
 
 func (k *Killgrave) getContainerRequest() (tc.ContainerRequest, error) {
-	killgraveImage, err := mirror.GetImage("friendsofgo/killgrave")
+	killgraveImage, err := mirror.GetImage("friendsofgo/killgrave:0.4.1")
 	if err != nil {
 		return tc.ContainerRequest{}, err
 	}


docker/test_env/killgrave.go
```diff
@@ -136,7 +136,7 @@ func (k *Killgrave) StartContainer() error {
 }
 
 func (k *Killgrave) getContainerRequest() (tc.ContainerRequest, error) {
-	killgraveImage, err := mirror.GetImage("friendsofgo/killgrave")
+	killgraveImage, err := mirror.GetImage("friendsofgo/killgrave:0.4.1")
```
tateexon, line 139: This should have been changed in mirror/mirror.json and not here or it will not update the mirror in ecr for when internal ecr usage is required

tateexon: Actually this looks like a big in the GetImage function since the mirror already has the version correct. We need a single source of truth and the mirror.json should be that.

tateexon: Testing the code out this morning now that I am back at my laptop, this change did not change the behavior.

tateexon:
```golang
	i, _ := GetImage("friendsofgo/killgrave")
	fmt.Println(i)
```
produces: friendsofgo/killgrave:0.4.1
and
```golang
	i, _ := GetImage("friendsofgo/killgrave:0.4.1")
	fmt.Println(i)
```
produces: friendsofgo/killgrave:0.4.1

tateexon: @lukaszcl where was this not producing the image with the version number?

lukaszcl: Hey @tateexon! Without this PR we we seeing ` unknown shorthand flag: 'o' in -ost=0.0.0.0` error. Please check [Slack](https://chainlink-core.slack.com/archives/C050QEMBFU5/p1700666731920129) for more details
 
```
      log.go:37: 00:07:45.60 INF container logs:
          2023/11/22 00:07:45 unknown shorthand flag: 'o' in -ost=0.0.0.0
          
          
      forwarder_ocr_test.go:30: 
          	Error Trace:	/home/runner/work/chainlink/chainlink/integration-tests/smoke/forwarder_ocr_test.go:30
          	Error:      	Received unexpected error:
          	            	container exited with code 1: failed to start container
          	            	cannot start Killgrave container
```

tateexon: So this change can safely be reverted and there isn't a bug in the mirror code that needs fixed.

lukaszcl: @tateexon ack, thanks for the info

PR #782, opened 2023-11-23
Add Nethermind, Erigon and Besu support

Tofel:
Also rework Geth support so that both use `ethereum genesis generator` and `eth2-val-tool` (which generates validator key).

After genesis/key generation we also use a small container to crate all shared files that don't require special logic:
* keystore
* keystore password
* wallet password


Tofel:
That code duplication is related to starting containers and in my opinion it is not worth to deal with it.

docker/test_env/nethermind.go
```diff
@@ -0,0 +1,203 @@
...
+	return &g.Container
+}
+
+func (g *Nethermind) getContainerRequest(networks []string) (*tc.ContainerRequest, error) {
+	return &tc.ContainerRequest{
+		Name:            g.ContainerName,
+		Image:           fmt.Sprintf("nethermind/nethermind:%s", NETHERMIND_IMAGE_TAG),
+		Networks:        networks,
+		AlwaysPullImage: true,
+		// ImagePlatform: "linux/x86_64", // this breaks everything, don't try it
```
skudasov, line 156: So, can it work on x86_64?

docker/test_env/validator_keys_generator.go
```diff
@@ -0,0 +1,100 @@
...
+
+func (g *ValKeysGeneretor) getContainerRequest(networks []string) (*tc.ContainerRequest, error) {
+	return &tc.ContainerRequest{
+		Name:          g.ContainerName,
+		Image:         "protolambda/eth2-val-tools:latest",
+		ImagePlatform: "linux/x86_64",
+		Networks:      networks,
+		WaitingFor: NewExitCodeStrategy().WithExitCode(0).
+			WithPollInterval(1 * time.Second).WithTimeout(10 * time.Second),
+		// Entrypoint: []string{"sh", "/init.sh"},
```
skudasov, line 80: Remove if not needed?

docker/test_env/genesis_generator_helpers.go
```diff
@@ -0,0 +1,287 @@
...
+	return w
+}
+
+func (w *ExitCodeStrategy) WithPollInterval(pollInterval time.Duration) *ExitCodeStrategy {
+	w.pollInterval = pollInterval
+	return w
+}
+
+// WaitUntilReady implements Strategy.WaitUntilReady
+func (w *ExitCodeStrategy) WaitUntilReady(ctx context.Context, target tcwait.StrategyTarget) (err error) {
```
skudasov, line 265: Seems like we are using it just to wait for the init container to finish in "success" state or something like that?

docker/docker.go
```diff
@@ -58,6 +58,10 @@ func StartContainerWithRetry(l zerolog.Logger, req tc.GenericContainerRequest) (
 				return nil, err
 			}
 		}
+		// TODO we should dynamically check if container was removed, otherwise retry will fail anyway, just becasuse
```
skudasov, line 61: I remember this was introduced because of Geth ports mapping issue, can we check if we can remove it now @AnieeG ?

skudasov (approved):

docker/test_env/nethermind.go
Tofel: Someone who has one should try, I only have Apple Silicon. But they do publish arm64 and x86 image versions (unlike other clients), so I imagine it should work.

docker/docker.go
Tofel: IMHO if we could remove it would be sweet, because I am yet to see a container that fails to start the first time, but then succeeds.

blockchain/ethereum.go
```diff
@@ -839,7 +839,12 @@ func (e *EthereumClient) EstimateGas(callMsg ethereum.CallMsg) (GasEstimations,
 		gasTipCap, err = e.Client.SuggestGasTipCap(ctx)
 		cancel()
 		if err != nil {
-			return GasEstimations{}, err
+			//For Besu, that support "eth_maxPriorityFeePerGas" https://github.com/hyperledger/besu/issues/5658
```
kalverra, line 842: Does Besu just make the estimation for us in this case @Tofel ?

docker/test_env/besu_test.go
```diff
@@ -0,0 +1,105 @@
...
+	require.NoError(t, err, "Couldn't start PoS network")
+
+	nonEip1559Network := blockchain.SimulatedEVMNetwork
+	nonEip1559Network.Name = "Simulated Besu + Prysm (non-EIP 1559)"
+	nonEip1559Network.GasEstimationBuffer = 10_000_000_000
+	nonEip1559Network.URLs = eth2.PublicWsUrls()
+	clientOne, err := blockchain.ConnectEVMClient(nonEip1559Network, l)
+	require.NoError(t, err, "Couldn't connect to the evm client")
+
+	defer func() {
```
kalverra, line 45: Why not use [t.Cleanup()](https://ieftimov.com/posts/testing-in-go-clean-tests-using-t-cleanup/)?

docker/test_env/erigon_test.go
```diff
@@ -0,0 +1,63 @@
...
+	_, eth2, err := cfg.Start()
+	require.NoError(t, err, "Couldn't start PoS network")
+
+	nonEip1559Network := blockchain.SimulatedEVMNetwork
+	nonEip1559Network.Name = "Simulated Erigon + Prysm (non-EIP 1559)"
+	nonEip1559Network.URLs = eth2.PublicWsUrls()
+	clientOne, err := blockchain.ConnectEVMClient(nonEip1559Network, l)
+	require.NoError(t, err, "Couldn't connect to the evm client")
+
+	defer func() {
```
kalverra, line 40: Same comment about t.Cleanup()

main
kalverra, line 1: Main binary checked in

docker/test_env/ethereum_env.go
```diff
@@ -68,28 +70,45 @@ func (b *EthereumNetworkBuilder) WithConsensusType(consensusType ConsensusType)
...
+		consensusLayer := ConsensusLayer_Prysm
+
+		b.participants = []EthereumNetworkParticipant{
+			{
+				ConsensusLayer: consensusLayer,
+				ExecutionLayer: ExecutionLayer_Geth,
+				Count:          1,
+			},
+		}
+	case ConsensusType_PoW:
```
kalverra, line 86: Do we want to even bother with this as an option anymore?

docker/test_env/ethereum_env.go
```diff
@@ -108,265 +127,363 @@ func (b *EthereumNetworkBuilder) WithTest(t *testing.T) *EthereumNetworkBuilder
...
+	isRecreated         bool
+	ethereumChainConfig *EthereumChainConfig
+	t                   *testing.T
+}
+
+func (en *EthereumNetwork) Start() (blockchain.EVMNetwork, RpcProvider, error) {
+	switch en.ConsensusType {
+	case ConsensusType_PoS:
+		return en.startPos()
+	case ConsensusType_PoW:
```
kalverra, line 236: Are there some tests that we still need this for? Or is this just to help along the transition?

docker/test_env/ethereum_env.go
```diff
@@ -108,265 +127,363 @@ func (b *EthereumNetworkBuilder) WithTest(t *testing.T) *EthereumNetworkBuilder
...
 	}
 
-	return net, RpcProvider{
-		privateHttpUrls: []string{docker.HttpUrl},
-		privatelWsUrls:  []string{docker.WsUrl},
-		publiclHttpUrls: []string{geth.ExternalHttpUrl},
-		publicsUrls:     []string{geth.ExternalWsUrl},
-	}, nil
+	en.DockerNetworkNames = networkNames
+	//TODO when we support multiple participants, we need to modify net so that it contains all the RPC URLs, not just the last one
```
kalverra, line 427: Is the support issue mainly within our test client, or with setting up the network?

docker/test_env/ethereum_env.go
```diff
@@ -441,38 +542,50 @@ func (e *EthereumNetworkContainers) wasAlreadyRestarted(containerName string) bo
...
-		if container.ContainerType == ct && !restartedContainers.wasAlreadyRestarted(container.ContainerName) {
-			restartedContainers.add(container)
-			return func(c *EnvComponent) {
-				if container.ContainerName != "" {
-					c.ContainerName = container.ContainerName
-				}
+	return customConfigDataDir, valKeysDir, nil
+}
+
+func waitForChainToFinaliseFirstEpoch(lggr zerolog.Logger, evmClient blockchain.EVMClient) error {
```
kalverra, line 559: How long does this usually take?

docker/test_env/genesis_generator.go
```diff
@@ -0,0 +1,157 @@
...
+		return nil, err
+	}
+	_, err = mnemonicsFile.WriteString(mnemonics)
+	if err != nil {
+		return nil, err
+	}
+
+	return &tc.ContainerRequest{
+		Name:          g.ContainerName,
+		Image:         "tofelb/ethereum-genesis-generator:2.0.4-slots-per-epoch",
```
kalverra, line 116: Need to ask about moving this to our ECR

kalverra (changes requested):

blockchain/ethereum.go
Tofel: Nope, we just set it to `0` and use just the base fee. But it won't work. I just tried Besu with contract upload and there we use `go-ethereum` client code to do the upload, which fails, because that method is not implemented. So we will need to work with Besu in non-EIP1559 mode until they have implemented `eth_maxPriorityFeePerGas` (or implement it ourselves in the upstream? ;-))

docker/test_env/besu_test.go
Tofel: true, hasn't occurred to me!

docker/test_env/ethereum_env.go
Tofel: good question, it's still much faster, when you're sending a lot of trxs. But what we could do here is for example run latest `Erigon` with `--fakepow` flag (I think Besu might support similar mode too).

docker/test_env/ethereum_env.go
Tofel: I didn't want to break backward-compatibility. Also, PoS is slower than PoW so some tests might fail because of that and we might to add some more checks/fluent waits.

docker/test_env/ethereum_env.go
Tofel: depends on the chain settings (seconds per slot + slots per epoch), with some reasonable ones (8 and 4 respectively) it might take around 1 - 1.5 minutes,

docker/test_env/genesis_generator.go
Tofel: I've opened a PR to the original repo, maybe they will merge it. If not, I'd create a private repo in our org, copy my fork there, then build a private image based on that and push it to our ECR. WDYT?

docker/docker.go
tateexon: Just chiming in here that I have seen this be helpful in CI in multiple cases, usually around the docker pull failing one of the times because of a server hiccup but then correctly pulls it the second or third time.

tateexon: Although we could add our own pull logic with retries so testcontainers-go doesn't have to do it.

docker/test_env/erigon.go
```diff
@@ -0,0 +1,257 @@
...
+		consensusLayer:       consensusLayer,
+		l:                    log.Logger,
+	}
+	for _, opt := range opts {
+		opt(&g.EnvComponent)
+	}
+	return g
+}
+
+func (g *Erigon) WithTestInstance(t *testing.T) *Erigon {
```
tateexon, line 63: Is there a reason we are not keeping the same naming consistency with existing containers? We have not been using WithTestInstance, we have been using WithTestLogger. We should keep consistent across the board for this kind of stuff, if we want to go with WithTestInstance (which I agree is a better name) then we need to either update all the others now or create a ticket to do it soon. We don't want to live with two patterns for too long.

Tofel: I changed it only, because it was a misleading name for me, because it was called `WithTestLogger` but we weren't passing logger, but `t`.

If you're okay with it, I'll change all the "old" methods also to the new one in this PR.

docker/test_env/ethereum_env.go
Tofel: But maybe we shouldn't make that wait the default? And instead explicitly requested?

docker/test_env/besu.go
```diff
@@ -0,0 +1,211 @@
...
+	tcwait "github.com/testcontainers/testcontainers-go/wait"
+
+	"github.com/smartcontractkit/chainlink-testing-framework/blockchain"
+	"github.com/smartcontractkit/chainlink-testing-framework/docker"
+	"github.com/smartcontractkit/chainlink-testing-framework/logging"
+	"github.com/smartcontractkit/chainlink-testing-framework/utils/testcontext"
+)
+
+const (
+	//TODO use Tate's mirror?
```
tateexon, line 24: I will make a ticket to create the ecr for these and get them added to the mirror and then update these in code.

docker/test_env/besu.go
```diff
@@ -0,0 +1,211 @@
...
+	InternalHttpUrl      string
+	ExternalWsUrl        string
+	InternalWsUrl        string
+	InternalExecutionURL string
+	ExternalExecutionURL string
+	generatedDataHostDir string
+	chainConfg           *EthereumChainConfig
+	consensusLayer       ConsensusLayer
+	l                    zerolog.Logger
+	t                    *testing.T
```
AnieeG, line 40: is it necessary to have *testing.T ? If someone wants to use it in some scripts other than tests, depending on *testing.T might not be ideal.

docker/test_env/besu.go
```diff
@@ -0,0 +1,211 @@
...
+func NewBesu(networks []string, chainConfg *EthereumChainConfig, generatedDataHostDir string, consensusLayer ConsensusLayer, opts ...EnvComponentOption) *Besu {
+	g := &Besu{
+		EnvComponent: EnvComponent{
+			ContainerName: fmt.Sprintf("%s-%s", "besu", uuid.NewString()[0:8]),
+			Networks:      networks,
+		},
+		chainConfg:           chainConfg,
+		generatedDataHostDir: generatedDataHostDir,
+		consensusLayer:       consensusLayer,
+		l:                    log.Logger,
```
AnieeG, line 52: does it make sense to pass the logger as parameter?

docker/test_env/besu.go
```diff
@@ -0,0 +1,211 @@
...
+}
+
+func (g *Besu) GetContainer() *tc.Container {
+	return &g.Container
+}
+
+func (g *Besu) getContainerRequest(networks []string) (*tc.ContainerRequest, error) {
+	return &tc.ContainerRequest{
+		Name:     g.ContainerName,
+		Image:    fmt.Sprintf("hyperledger/besu:%s", BESU_IMAGE_TAG),
```
AnieeG, line 157: might be helpful if we have a method like `WithImage`

docker/test_env/besu.go
```diff
@@ -0,0 +1,211 @@
...
+				Source: tc.GenericBindMountSource{
+					HostPath: g.generatedDataHostDir,
+				},
+				Target: tc.ContainerMountTarget(GENERATED_DATA_DIR_INSIDE_CONTAINER),
+			},
+		},
+	}, nil
+}
+
+func (g Besu) WaitUntilChainIsReady(waitTime time.Duration) error {
```
AnieeG, line 204: Does it give any warning if you have method on both pointer and non-pointer object?

docker/test_env/erigon.go
```diff
@@ -0,0 +1,257 @@
...
+	InternalHttpUrl      string
+	ExternalWsUrl        string
+	InternalWsUrl        string
+	InternalExecutionURL string
+	ExternalExecutionURL string
+	generatedDataHostDir string
+	chainConfg           *EthereumChainConfig
+	consensusLayer       ConsensusLayer
+	l                    zerolog.Logger
+	t                    *testing.T
```
AnieeG, line 43: same concern

docker/test_env/besu.go
Tofel: unfortunately it is, re: thread in Slack

docker/test_env/besu.go
Tofel: you tell me :-) what would be the use case? unless there's a strong one I'd rather not bloat the constructor too much. Or... I could add a functional option for passing logger?

Also maybe by default we should initialise the logger using `logging.GetTestLogger(t)` even if `t` is `nil` at that point?

docker/test_env/besu.go
Tofel: good catch, there's no reason not to use pointer receiver

docker/test_env/besu.go
Tofel: @AnieeG for now I've implemented the latter

docker/test_env/erigon.go
Tofel: @tateexon I renamed everything to `WithTestInstance`

lukaszcl: +1 for `WithTestInstance` name as testing.T is passed there. But, @Tofel do we really need whole `*testing.T`. If this is only for logging maybe you can initialize the logger with `testing.T`?

docker/docker.go
lukaszcl: I agree with @tateexon here about custom pulling logic instead of retrying whole `tc.GenericContainer()`. Here's how testcontainers-go [does it](https://github.com/testcontainers/testcontainers-go/blob/main/docker.go#L1175). @Tofel but, the pulling logic with retry can be a separate ticket

docker/test_env/besu.go
```diff
@@ -0,0 +1,221 @@
...
+		l:                    logging.GetTestLogger(nil),
+		image:                fmt.Sprintf("hyperledger/besu:%s", BESU_IMAGE_TAG),
+	}
+	for _, opt := range opts {
+		opt(&g.EnvComponent)
+	}
+	return g
+}
+
+func (g *Besu) WithImage(imageWithTag string) *Besu {
```
lukaszcl, line 61: @Tofel maybe you could move it to `EnvComponent` and rename to `WithContainerImage()`? Same can be done with `WithContainerVersion`

docker/test_env/besu.go
```diff
@@ -0,0 +1,221 @@
...
+	return g.ExternalWsUrl
+}
+
+func (g *Besu) GetContainerName() string {
+	return g.ContainerName
+}
+
+func (g *Besu) GetContainer() *tc.Container {
+	return &g.Container
+}
```
lukaszcl, line 162: @Tofel maybe `GetContainerName()` and `GetContainer()` can be moved to `EnvComponent`?

docker/test_env/besu.go
```diff
@@ -0,0 +1,221 @@
...
+				},
+				Target: tc.ContainerMountTarget(GENERATED_DATA_DIR_INSIDE_CONTAINER),
+			},
+		},
+	}, nil
+}
+
+func (g *Besu) WaitUntilChainIsReady(waitTime time.Duration) error {
+	waitForFirstBlock := tcwait.NewLogStrategy("Imported #1").WithPollInterval(1 * time.Second).WithStartupTimeout(waitTime)
+	return waitForFirstBlock.WaitUntilReady(context.Background(), *g.GetContainer())
```
lukaszcl, line 216: @Tofel shouldn't `testcontext.Get` be used here?

lukaszcl (commented):
@Tofel I added few comments that apply to all docker wrappers/components. Please check. Otherwise looks really good to me!

docker/test_env/besu.go
Tofel: that was my question in priv to @tateexon actually, I think it should. Changed it and also changed all calls to `context.Background()` in `test_env` package to that one

docker/test_env/besu.go
Tofel: yeah, but can we do it in a separate PR? this one is already biig

docker/test_env/besu.go
Tofel: yeah, but can we do it in a separate PR? this one is already biig

docker/test_env/besu.go
Tofel: we need to pass `*testing.T` anyway :/

docker/test_env/besu.go
lukaszcl: yes

docker/test_env/besu.go
lukaszcl: yes

Tofel: Added to https://smartcontract-it.atlassian.net/browse/TT-743

PR #783, opened 2023-11-23
Add Erigon support

Tofel:


Tofel:
code duplication is related to starting containers, imho it doesn't make sense to fix as reduced duplication would be low, but complexity much higher than now

skudasov (approved):

PR #784, opened 2023-11-24
Add Besu support

Tofel:
* handle lack of implemented `eth_maxPriorityFeePerGas` method in RPC, when using EIP-1559 (case of Besu's new tx pool)
* update all tests, so that they run with EIP-1559 network and without it


docker/test_env/ethereum_env.go
```diff
@@ -354,21 +372,10 @@ func (en *EthereumNetwork) startPos() (blockchain.EVMNetwork, RpcProvider, error
 	}
 
 	en.DockerNetworkNames = networkNames
-	//TODO when we support multiple participants, we need to modify net so that it contains all the RPC URLs, not just the last one
+	net.ChainID = int64(en.ethereumChainConfig.ChainID)
+	net.SupportsEIP1559 = true
```
skudasov, line 376: So all nets do support EIP1559 by default, except Besu? We turn it on in `ethereum_env.go` by default but I don't see where we set this flag off for Besu. If we skip executing `eth_maxPriorityFeePerGas` if client type is `Besu` and never set this flag to `false` maybe we can remove it?

docker/test_env/wait_strategies.go
```diff
@@ -0,0 +1,108 @@
...
+	// the number of occurrence needs to be positive
+	if o <= 0 {
+		o = 1
+	}
+	ws.Occurrence = o
+	return ws
+}
+
+// WaitUntilReady implements Strategy.WaitUntilReady
+func (ws *LogRegexStrategy) WaitUntilReady(ctx context.Context, target tcwait.StrategyTarget) (err error) {
```
skudasov, line 49: Can we add a comment about what we are searching for and why it's so specific?

skudasov (approved):

docker/test_env/wait_strategies.go
Tofel: I think I will remove it, because I found a better way to make sure that an epoch is finalized.

docker/test_env/ethereum_env.go
Tofel: Besu also supports EIP1559, but not our implementation of it's handling, i.e. it works with MetaMask, so I guess they handle a situation, when that very method is not implemented.

With the small change to `blockchain/ethereum.go` that I did in this PR we also handle that and trx are correctly gas-priced and processed.

PR #802, opened 2023-12-15
[BREAKING] TOML config

Tofel:
# Intro

Adds TOML-based config for following elements:

* `LogStream`
* `Selected networks`
* `Private Ethereum Network`
* `Pyroscope`

Based on what @AnieeG and @skudasov did in CCIP/PoC I also introduced a slightly modified interface that all configs should implement, but I'm not really using it anywhere. Not sure if it's really needed other than in some descriptive way to say what methods are expected. I was thinking about using it together with composite pattern or reflection package in order to iterate over all struct fields, see if they implement the interface and if they do call some of the methods (like `ApplyOverrides()`), which would hugely reduce the boilerplate code, but in the end decided against it. Somehow I get this feeling that using reflection in Go is a bit of no-go, if it was Java/Kotlin I'd do it. But if you think that would be a better solution, do let me know.

On purpose and contrary to what @AnieeG did for CCIP I did not include `Read()` or `ReadSecrets()` methods as I thought it would introduce too much work on the implementation front. I decided it will be easier to construct the final test config from the building blocks present in CTF and product-specific ones and then read top-level config and secrets all at once instead of doing it in iterative manner for reach sub-config.

The only exception to it is `NetworkConfig`, which has `ApplySecrets()` function, which can be used to read default RPC and private keys configuration. The idea is to put that information in TOML, base64 it and then store it as secret in GHA. Then when tests run we'd grab it from secrets and expose as env var. That could save us the need to provide URLs every time we run a job on a live testnet (especially if we can come up with a set of reliable ones, but I think we do have that since @kalverra live testnet job must be already using some predefined RPC set).

# Main changes
## Recursive file-walker

I also added a utility function to recursively walk through directories until a desired file is found (with a configurable limit of parent directories so that we don't end up traversing whole filesystem). The idea behind is to help us localise TOML files inside the repo regardless of current working directory. First in the top-most directory that should be accessed we place `.root_dir` file which will be looked for first, once it's found starting from there the code will walk through all directories trying to find the config we want. It will return first file it finds.

## CL image/version override in Helm

I added a new method `NewWithOverride()` to `config` package, so that we can override CL image/version from TOML config instead of env vars, when building Helm charts. The old `New()` method is still there and still uses env vars for override.

## Ethereum env

I removed the ability to set execution client from env var, now TOML config has to be used for that too.

## Remote runner

I added automatic forwarding of these 2 env vars to remote runner:
* `BASE64_CONFIG_OVERRIDE`
* `BASE64_NETWORK_CONFIG`

# Breaking changes
* `SELECTED_NETWORK` env variable is no longer
  * `func MustGetSelectedNetworksFromEnv() []blockchain.EVMNetwork` --> 
`func MustGetSelectedNetworkConfig(networkCfg *config.NetworkConfig) []blockchain.EVMNetwork`
  * `func SetNetworks(networkKeys []string) []blockchain.EVMNetwork` --> `func MustSetNetworks(networkCfg config.NetworkConfig) []blockchain.EVMNetwork`
* LogStream no longer accepts some functional options, instead it takes them from logging config
* Pyroscope is no longer configured via env var
  * `func AddNetworksConfig(baseTOML string, networks ...blockchain.EVMNetwork) string` --> `func AddNetworksConfig(baseTOML string, pyroscopeConfig *config.PyroscopeConfig, networks ...blockchain.EVMNetwork) string`
  * `func AddNetworkDetailedConfig(baseTOML string, detailedNetworkConfig string, network blockchain.EVMNetwork) string` -- > `func AddNetworkDetailedConfig(baseTOML string, pyroscopeConfig *config.PyroscopeConfig, detailedNetworkConfig string, network blockchain.EVMNetwork) string`
* `JSONStrDuration` --> `StrDuration` (added TOML (un)marshaller so the name wasn't adequate anymore)
* instead of using `chainlink.New()` when building Helm chart you should now use `chainlink.NewWithOverride()` which accepts two new parameters: `testconfig.ChainlinkImage` and custom `overrideFn`, which tells it how to override CL image and version from TOML instead of env vars. That override function has to be very specific (I'm using it as a closure, so that it captures `testconfig` from the scope). Same applies to setting Pyroscope auth key. Example to follow:
```
import ctf_config "github.com/smartcontractkit/chainlink-testing-framework/config"

config, err := tc.GetConfig(testType, tc.Keeper)
require.NoError(t, err, "Error getting test config")

var overrideFn = func(_ interface{}, target interface{}) {
	ctf_config.MustConfigOverrideChainlinkVersion(config.ChainlinkImage, target)
        ctf_config.MightConfigOverridePyroscopeKey(config.Pyroscope, target)
}

cd := chainlink.NewWithOverride(i, map[string]any{
	"toml":      networks.AddNetworkDetailedConfig(keeperBenchmarkBaseTOML, config.Pyroscope, networkDetailTOML, testNetwork),
	"chainlink": chainlinkResources,
	"db":        dbResources,
}, config.ChainlinkImage, overrideFn)
```

# Not done
* Slack-related configuration is still env-var based (do we want to move it to TOML as well?)

# Usage
Core changes have been applied in [this PR](https://github.com/smartcontractkit/chainlink/pull/11588).


utils/osutil/osutil.go
```diff
@@ -96,3 +98,81 @@ func GetAbsoluteFolderPath(folder string) (string, error) {
...
+
+const (
+	DEFAULT_STOP_FILE_NAME         = ".root_dir"
+	ErrStopFileNotFoundWithinLimit = "stop file not found in any parent directory within limit"
+)
+
+// FindFile looks for given file in the current directory and its parent directories first by locating
+// top level parent folder where the search should begin (defined by stopFile, which cannot be located
+// further "up" than limit parent folders) and then by searching from there for the file all subdirectories
+func FindFile(filename, stopFile string, limit int) (string, error) {
```
tateexon, line 110: Should we add a test for this function?

k8s/config/overrides.go
```diff
@@ -98,25 +92,21 @@ const (
...
 	EnvVarLocalChartsUserDescription = "Use local charts from the CTF repository directly"
 	EnvVarLocalChartsExample         = "true"
+
+	EnvBase64ConfigOverride             = "BASE64_CONFIG_OVERRIDE"
+	EnvBase64ConfigOverriderDescription = "Base64-encoded TOML config (should contain at least chainlink image and version)"
+	EnvBase64ConfigOverrideExample      = "W0NoYWlubGlua0ltYWdlXQppbWFnZT0icHVibGljLmVjci5hd3MvY2hhaW5saW5rL2NoYWlubGluayIKdmVyc2lvbj0iMi43LjEtYXV0b21hdGlvbi0yMDIzMTEyNyIKCltBdXRvbWF0aW9uXQpbQXV0b21hdGlvbi5HZW5lcmFsXQpkdXJhdGlvbj0yMDAK"
+
+	EnvBase64NetworkConfig            = "BASE64_NETWORK_CONFIG"
+	EnvBase64NetworkConfigDescription = "Base64-encoded TOML network config with default RPC/WS endpoints and private keys"
+	EnvBase64NetworkConfigExample     = "Nope :-)"
```
tateexon, line 109: lol

docker/test_env/ethereum_env.go
```diff
@@ -11,6 +11,8 @@ import (
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/google/uuid"
+	"github.com/pelletier/go-toml/v2"
+	"github.com/pkg/errors"
```
tateexon, line 15: I thought we got rid of usages of this since it is no longer maintained. We had replaced most usages with fmt.Errorf

docker/test_env/eth2_common.go
```diff
@@ -2,10 +2,14 @@ package test_env
...
 import (
 	"context"
+	_ "embed"
 	"fmt"
 	"testing"
 	"time"
 
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/pelletier/go-toml/v2"
+	"github.com/pkg/errors"
```
tateexon, line 12: Another usage of a no longer maintained library. Preferable to use fmt.Errorf for now.

config/pyroscope.go
```diff
@@ -0,0 +1,56 @@
+package config
+
+import (
+	"github.com/pkg/errors"
```
tateexon, line 4: Preferable to use fmt.Errorf for now instead.

config/network.go
```diff
@@ -0,0 +1,159 @@
+package config
+
+import (
+	"encoding/base64"
+	"strings"
+
+	"github.com/pelletier/go-toml/v2"
+	"github.com/pkg/errors"
```
tateexon, line 8: Another usage.

config/logging.go
```diff
@@ -0,0 +1,184 @@
+package config
+
+import (
+	_ "embed"
+
+	"github.com/pelletier/go-toml/v2"
+	"github.com/pkg/errors"
```
tateexon, line 7: Another usage

tateexon (changes requested):

docker/test_env/ethereum_env.go
tateexon: For reference we had removed all references in this PR back in November: https://github.com/smartcontractkit/chainlink-testing-framework/pull/762

utils/osutil/osutil.go
Tofel: would probably make a lot of sense

k8s/config/overrides.go
Tofel: hihi I thought it wouldn't add to much value to add random base64ed TOML, but if you think it makes sense I can do it

.nancy-ignore
```diff
@@ -9,4 +9,4 @@ CVE-2022-29153 # Server-Side Request Forgery (SSRF)
 CVE-2023-3518 # CWE-Other (when using JWT Auth for service mesh incorrectly allows/denies access regardless of service identities) -- coming from WASP
 CVE-2023-40591 # CWE-400 Uncontrolled Resource Consumption ('Resource Exhaustion') by geth (fixed in v1.12.2, which needs core to bump that dep first)
 CVE-2023-42319 # CWE-400 Uncontrolled Resource Consumption ('Resource Exhaustion') by geth (fixed in v1.12.2, which needs core to bump that dep first)
-CVE-2023-48795 # Improper Validation of Integrity Check Value x/crypto, which is an indirect dependency
\ No newline at end of file
+CVE-2023-48795 # Improper Validation of Integrity Check Value x/crypto, hich is an indirect dependency
```
AnieeG, line 12: spell check

README.md
```diff
@@ -48,6 +49,16 @@ go run k8s/examples/simple/env.go
 ```
 For more features follow [tutorial](./k8s/TUTORIAL.md)
 
+#### With TOML config
+It should be noted that using env vars for configuring CL nodes in k8s is deprecated. TOML config should be used instead:
+```toml
+[ChainlinkImage]
+image="public.ecr.aws/chainlink/chainlink"
```
AnieeG, line 56: can it be considered as secret?

config/README.md
```diff
@@ -0,0 +1,156 @@
...
+
+It is advised to add `overrides.toml` to `.gitignore`.
+
+## Network config (and default RPC endpoints)
+
+Some more explanation is needed for the `NetworkConfig`:
+```golang
+type NetworkConfig struct {
+	SelectedNetworks []string            `toml:"selected_networks"`
+	RpcHttpUrls      map[string][]string `toml:"RpcHttpUrls"`
```
AnieeG, line 48: please provide a comment on what will be the keys for these map

config/examples/testconfig.go
```diff
@@ -0,0 +1,238 @@
...
+	if c.PrivateEthereumNetwork != nil {
+		if err := c.PrivateEthereumNetwork.Validate(); err != nil {
+			return fmt.Errorf("private ethereum network config validation failed: %w", err)
+		}
+	}
+
+	return nil
+}
+
+func readFile(filePath string) ([]byte, error) {
```
AnieeG, line 231: move it some common utils?

config/network.go
```diff
@@ -0,0 +1,161 @@
...
+
+		if _, ok := n.RpcHttpUrls[network]; !ok {
+			return fmt.Errorf("At least one HTTP RPC endpoint for %s network must be set", network)
+		}
+
+		if _, ok := n.RpcWsUrls[network]; !ok {
+			return fmt.Errorf("At least one WS RPC endpoint for %s network must be set", network)
+		}
+
+		if _, ok := n.WalletKeys[network]; !ok {
```
AnieeG, line 113: how will we deal with keys which we don't want to know and want them to be managed by security? cc @kalverra

config/network.go
```diff
@@ -0,0 +1,161 @@
...
+			for network, urls := range from.WalletKeys {
+				n.WalletKeys[network] = urls
+			}
+		}
+	}
+
+	return nil
+}
+
+func (n *NetworkConfig) Default() error {
```
AnieeG, line 159: Should we add default from env vars?

docker/test_env/ethereum_env.go
skudasov: There is an ongoing conversation, we haven't found an alternative yet, so I'm okay to keep it.

README.md
Tofel: could you expand on that? why would we consider image name a secret? in this very case this is a public ecr repo, so not much of a secret.

In GHA in all jobs that use images stored on our private ECR we should mask make the image (and in core that's what we do in all jobs)

config/examples/testconfig.go
Tofel: don't you think it's too small of a method to move it to utils?

config/network.go
Tofel: could you expand on that?

what I'm doing now in live testnet job is this:
* that job takes EVM keys from GH secrets (so GH automatically masks it in log output)
* I build base64 config override with network details, including the private kay and mask it with `::add-mask::`
* I set that override as env var that tests read

Do you see any potential risk with that approach?

config/network.go
Tofel: Do you mean making the `Default()` actually execute [this method](https://github.com/smartcontractkit/chainlink-testing-framework/pull/802/files/85ab90ffd3b5456bad3f1477d60fcb07014d9067#diff-32f4f366dad3cee33f9165dbf93cf213721a7dd734bf3e99cd295e35ba122e4dR26)? Or something yet different?

config/network.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	for _, network := range n.SelectedNetworks {
+		if strings.Contains(network, "SIMULATED") {
+			// we don't need to validate RPC endpoints or private keys for simulated networks
+			continue
+		}
+
+		if _, ok := n.RpcHttpUrls[network]; !ok {
+			return fmt.Errorf("At least one HTTP RPC endpoint for %s network must be set", network)
```
skudasov, line 106: nitpick, lowercase for all errors, 2 more in this function

config/network.go
```diff
@@ -0,0 +1,161 @@
...
+func (n *NetworkConfig) Validate() error {
+	if len(n.SelectedNetworks) == 0 {
+		return errors.New("selected_networks must be set")
+	}
+
+	upperCasedHttpUrls := make(map[string][]string)
+	for key := range n.RpcHttpUrls {
+		upperCasedHttpUrls[strings.ToUpper(key)] = n.RpcHttpUrls[key]
+	}
+	n.RpcHttpUrls = upperCasedHttpUrls
```
skudasov, line 81: So `Validate()` is mutating our config, can't it cause any issues with URLs?

skudasov: Ah, it's the key, then okay

config/logging.go
```diff
@@ -0,0 +1,203 @@
...
+	if l.Grafana != nil {
+		if err := l.Grafana.Validate(); err != nil {
+			return fmt.Errorf("invalid grafana config: %w", err)
+		}
+	}
+
+	return nil
+}
+
+func (l *LoggingConfig) ApplyOverrides(from *LoggingConfig) error {
```
skudasov, line 41: We can do simple multi-marshal to the same pointer, example
```
package main

import (
	"github.com/davecgh/go-spew/spew"
	"github.com/pelletier/go-toml/v2"
	"os"
)

type Config struct {
	Group1 *Group1 `toml:"group_1"`
}

type Group1 struct {
	Name         string  `toml:"name"`
	Surname      string  `toml:"surname"`
	UniqueWeapon string  `toml:"unique_weapon"`
	Inventory    []int64 `toml:"inventory"`
}

func readAndOverride(prevCfg *Config, name string) (*Config, error) {
	f, err := os.ReadFile(name)
	if err != nil {
		panic(err)
	}
	if err := toml.Unmarshal(f, &prevCfg); err != nil {
		panic(err)
	}
	spew.Dump(prevCfg)
	return prevCfg, nil
}

func main() {
	cfg := &Config{}
	cfg1, err := readAndOverride(cfg, "config_base.toml")
	if err != nil {
		panic(err)
	}
	cfg2, err := readAndOverride(cfg1, "config_product.toml")
	if err != nil {
		panic(err)
	}
	_, err = readAndOverride(cfg2, "config_override.toml")
	if err != nil {
		panic(err)
	}
}
```
Output
```
(*main.Config)(0x1400005a038)({
 Group1: (*main.Group1)(0x1400007a140)({
  Name: (string) (len=4) "John",
  Surname: (string) (len=4) "Snow",
  UniqueWeapon: (string) (len=5) "sword",
  Inventory: ([]int64) (len=2 cap=16) {
   (int64) 1,
   (int64) 2
  }
 })
})
(*main.Config)(0x1400005a038)({
 Group1: (*main.Group1)(0x1400007a140)({
  Name: (string) (len=4) "John",
  Surname: (string) (len=4) "Snow",
  UniqueWeapon: (string) (len=5) "sword",
  Inventory: ([]int64) (len=1 cap=16) {
   (int64) 3
  }
 })
})
(*main.Config)(0x1400005a038)({
 Group1: (*main.Group1)(0x1400007a140)({
  Name: (string) (len=4) "John",
  Surname: (string) (len=9) "Malkovich",
  UniqueWeapon: (string) (len=5) "sword",
  Inventory: ([]int64) (len=1 cap=16) {
   (int64) 999
  }
 })
})
```
That can reduce amount of logic in cfg code we need to write and maintain, and overriding this way is more straightforward

config/config.go
```diff
@@ -0,0 +1,8 @@
+package config
+
+// GenericConfig is an interface for all product based config types to implement
+type GenericConfig[T any] interface {
+	Validate() error
+	ApplyOverride(from T) error
```
skudasov, line 6: Should we add `ApplySecrets()` to the interface so we can put all secrets handling in these methods?

config/logging.go
skudasov: This applies to all config structures

skudasov (changes requested):
LGTM, but I'd like to check if we can just marshal multiple times and remove all "default" overriding code. It's okay to use `ApplyOverrides()` but only in corner cases where you don't work with files or structs, like a matrix job in CI or env vars/secrets. All the other attempts to override configs can be done with simple marshalling.

README.md
AnieeG: There can be scenarios where in on-demand jobs, one might want to run a test against specific image input. For example I sometime run test with specific ECR image which is not stored in QA ECR but in production ECR. In that case masking the image becomes necessary.

config/network.go
AnieeG: My question was whether we could use EVM Keys from github secrets, if we can do that, my question is answered.

config/network.go
AnieeG: yes

config/config.go
Tofel: I actually don't think we need to keep this interface.

Tofel:
> LGTM, but I'd like to check if we can just marshal multiple times and remove all "default" overriding code. It's okay to use `ApplyOverrides()` but only in corner cases where you don't work with files or structs, like a matrix job in CI or env vars/secrets. All the other attempts to override configs can be done with simple marshalling.

I think it's a good idea, I tried it and it works.

Now I'm wondering how to make it less likely to leak secrets. Maybe this would be safer than what we have now:
```
type Config struct {
	Secret *string `toml:"secret"`
}

func (c *Config) ReadSecret() error {
	secret, isSet := os.LookupEnv("SECRET")
	if !isSet {
		return nil
	}

	c.Secret = &secret

	return nil
}
```

There would be nothing stopping you from doing this:
```
[Config]
my_secret="blabla"
```

but it would be less likely?

I think that might be more secure, but it would a bit fragment the configuration in the sense that some things would be set from TOML and other from env vars.

What do you think?

config/network.go
Tofel: you are right, I will move it to dedicated method, my idea was that we shouldn't really care whether the users input network as `Polygon_Mumbai` or `polygon_mumbai` or `POlygoN_MumbAI`, but I think it's better to move that code to sth like `UpperCaseNetworkNames()`

skudasov:
> > LGTM, but I'd like to check if we can just marshal multiple times and remove all "default" overriding code. It's okay to use `ApplyOverrides()` but only in corner cases where you don't work with files or structs, like a matrix job in CI or env vars/secrets. All the other attempts to override configs can be done with simple marshalling.
> 
> I think it's a good idea, I tried it and it works.
> 
> Now I'm wondering how to make it less likely to leak secrets. Maybe this would be safer than what we have now:
> 
> ```
> type Config struct {
> 	Secret *string `toml:"secret"`
> }
> 
> func (c *Config) ReadSecret() error {
> 	secret, isSet := os.LookupEnv("SECRET")
> 	if !isSet {
> 		return nil
> 	}
> 
> 	c.Secret = &secret
> 
> 	return nil
> }
> ```
> 
> There would be nothing stopping you from doing this:
> 
> ```
> [Config]
> my_secret="blabla"
> ```
> 
> but it would be less likely?
> 
> I think that might be more secure, but it would a bit fragment the configuration in the sense that some things would be set from TOML and other from env vars.
> 
> What do you think?

You can have TOML as `base64` or even raw in this env var, why not? The way how we pull secrets should be separated from the public configuration because it may change in the future and it should not affect the configuration.

I think that we should use `before-commit` hook with some software to prevent the leakage, except that I don't see any secure ways to prevent leakage if it's commited or if we missed possible secret logging on review, doesn't matter where you put it, you can still dump it or misuse mask in GitHub so:
- Separate the secrets reading logic
- Use `pre-commit` hook to find secrets in all the places
- Maybe some tool can help us to understand if we log something that can be secret

config/network.go
Tofel: I modified it so that what `Defaults()` actually does is read env var

README.md
Tofel: but if our AWS account or region is secret GH will mask them anyway in logs, right?

utils/osutil/osutil.go
Tofel: added

AnieeG (approved):

skudasov (approved):

gheorghestrimtu (approved):

gheorghestrimtu (approved):

go.mod
```diff
@@ -71,7 +95,7 @@ replace (
 )
 
 require (
-	dario.cat/mergo v1.0.0 // indirect
+	dario.cat/mergo v1.0.0
```
tateexon, line 98: Move this up to be with the direct deps and not down with the indirects. Typically this is done automatically unless this dep was changed by hand.

tateexon (changes requested):

tateexon (approved):

PR #804, opened 2023-12-19
[charts] Besu in k8s

Tofel:
* defines a Helm package `eth2-common` with reusable charts
* adds Besu execution client to k8s
* refactors previous eth2 Geth charts to use the common package

left to do:
* update dependency location to URL
* see if there's a better way to define storage


ajgrande924:
@Tofel I just noticed that youre provisioning deployments and creating a pv and pvc. What if you switch to a statefulset and define the [volumeClaimTemplates](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#volume-claim-templates)? Also, I think you could do this w/ readwriteonce and deploy all containers that write to the same storage as either an init container or as a sidecar container. You can proally then just use a regular ebs volume. If you do all this, you do not need to define the pv and your hostpath problem goes away (your pv should be created dynamically, you will just need a pvc w/ a filesystem volumemode). When you manually create the pv and define the hostpath, youre using the filesystem of the k8s node, Im not sure you want to do this, this is usually done for perf improvements b/c your k8s nodes have some local ssd attached to it that you want to access. Lemme know what you think. It may be a good idea to get inspiration from 3rd party charts or just use those under the hood.

ajgrande924:
@Tofel some inspiration https://github.com/ethpandaops/ethereum-helm-charts

PR #811, opened 2024-01-10
[TT-755] grafana url shortener

Tofel:
LogStream will try to shorten grafana urls, when:
* grafana url is set **AND**
* dashboard url is set **AND**
* bearer token is set

Otherwise it will return partial url (e.g. if base url is not set), full url (if token is not set) or just query parameters (if neither base url nor dashboard url is set).

If shoretning fails original url will be returned.

### DIFF
--- a/logstream/grafana.go
+++ b/logstream/grafana.go
@@ -1,0 +1,43 @@
+package logstream
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"net/http"
+)
+
+const (
+	ShorteningFailedErr = "failed to shorten Grafana URL"
+)
+
+func ShortenUrl(grafanaUrl, urlToShorten, bearerToken string) (string, error) {
+	jsonBody := []byte(`{"path":"` + urlToShorten + `"}`)
+	bodyReader := bytes.NewReader(jsonBody)
+
+	var responseObject struct {
+		Uid string `json:"uid"`
+		Url string `json:"url"`
+	}
+
+	req, err := http.NewRequest(http.MethodPost, grafanaUrl, bodyReader)
+	if err != nil {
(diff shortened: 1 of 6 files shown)


logstream/logstream_user_loki_test.go
```diff
@@ -53,12 +53,33 @@ func TestExampleLokiStreaming(t *testing.T) {
...
 
 			lw, err := logstream.NewLogStream(t, &loggingConfig)
 			require.NoError(t, err)
-			err = d.ConnectLogs(lw)
-			require.NoError(t, err)
+			for _, c := range d.containers {
+				err = lw.ConnectContainer(ctx, c, "")
+				require.NoError(t, err)
+			}
 			time.Sleep(5 * time.Second)
```
kalverra, line 72: Just seeing this now, why the hard sleep? Can we make it dynamic?

logstream/grafana.go
```diff
@@ -0,0 +1,47 @@
...
+	"encoding/json"
+	"fmt"
+	"net/http"
+)
+
+const (
+	ShorteningFailedErr = "failed to shorten Grafana URL"
+)
+
+func ShortenUrl(grafanaUrl, urlToShorten, bearerToken string) (string, error) {
```
kalverra, line 14: Can we add some unit tests for this?

kalverra (changes requested):

logstream/logstream_user_loki_test.go
Tofel: Ah, this sleep is here so that the test container we have started earlier have time to emit some logs before test ends :-)

logstream/grafana.go
Tofel: I think that would be hard/overkill as we hit actual API here

kalverra (approved):

PR #812, opened 2024-01-10
[TT-769]  customisable docker images

Tofel:
Now you will be able to override default Docker images for all components used, when creating a private Ethereum chain. hat works both, when using builder and when starting the chain via command line.

Closes https://smartcontract-it.atlassian.net/browse/TT-769


utils/toml/toml.go
```diff
@@ -0,0 +1,43 @@
...
+
+import (
+	"errors"
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/pelletier/go-toml/v2"
+)
+
```
tateexon, line 11: Would it make sense to add a some unit tests for these?

Tofel: added

tateexon (approved):

AnieeG (approved):

PR #813, opened 2024-01-11
[TT-792] Makes 'file' Default Log Target

kalverra:
Also lowers logging level for most log collection actions as we were getting a lot of noise.

### DIFF
--- a/logstream/logstream.go
+++ b/logstream/logstream.go
@@ -841,23 +841,22 @@ func (g *ContainerLogConsumer) hasLogTarget(logTarget LogTarget) bool {
 // getLogTargetsFromEnv gets log targets from LOGSTREAM_LOG_TARGETS env var
 func getLogTargetsFromEnv() ([]LogTarget, error) {
 	envLogTargetsValue := os.Getenv("LOGSTREAM_LOG_TARGETS")
-	if envLogTargetsValue != "" {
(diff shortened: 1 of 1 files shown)


logstream/logstream.go
```diff
@@ -855,9 +855,9 @@ func getLogTargetsFromEnv() ([]LogTarget, error) {
 				return []LogTarget{}, errors.Errorf("unknown log target: %s", target)
 			}
 		}
-
 		return envLogTargets, nil
 	}
 
-	return []LogTarget{}, nil
+	// default log target is file
+	return []LogTarget{"file"}, nil
```
Tofel, line 862: What do you think of moving it out of this method, to where it's called, and adding `file` only if what env returns is an empty array?

kalverra: I thought that's what this would be doing? We check if it's empty or not [here](https://github.com/smartcontractkit/chainlink-testing-framework/pull/813/files#diff-c6fa5a005831096c282975fd746aab1e8b7730648d1ac4502e3a7bce8d8134f7R844), and if it is, we'd go down to this return, right?

kalverra: Changed flow of function for I think more clear behavior

Tofel (approved):

PR #816, opened 2024-01-16
Add LogStream to EnvComponent

Tofel:
and... stealing @lukaszcl's work define defaults post start/end hooks that connect/disconnect logs if log stream is set.


AnieeG (approved):
Perfect

PR #817, opened 2024-01-16
[TT-714] Customisable container start retrying

Tofel:
For some images we have a situation, when image download fails, because there's no image for our platform (that's the case of CLL image, when you're running on Apple Silicon).  Error message is confusing.

Originally I thought that I could just add `linux/x86_84` image platform to container request, but then I thought that might not be optimal to hard code it.

So instead I added some strategies to retry starting of container. I wanted to do it using chain of responsibility pattern, but that proved impossible without changing method's signature and without having an interface implemented by specific types representing different strategies. So I opted for a simple approach, where each retry strategy is just a function with the same signature. We can pass a vararg of these functions when starting a container and we will try to use all of them in given order until one of them succeeds or we use all retry attempts.

### DIFF
--- a/docker/docker.go
+++ b/docker/docker.go
@@ -2,6 +2,7 @@ package docker
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/google/uuid"
 	"github.com/rs/zerolog"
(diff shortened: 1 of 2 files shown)


docker/docker.go
```diff
@@ -42,27 +43,117 @@ func CreateNetwork(l zerolog.Logger) (*tc.DockerNetwork, error) {
...
+		Str("Current start error", err.Error()).
+		Str("Retrier", "PlatoformImageRetrier").
+		Msgf("Failed to start %s container,", req.Name)
+
+	return nil, startErr
+}
+
+// StartContainerWithRetry attempts to start a container with 3 retry attempts.
+// It will try to start the container with the provided retriers, if none are provided it will use the default retriers.
+// Default being: 1. tries to download image for "linux/x86_64" platform 2. simply starts again without changing anything
```
lukaszcl, line 133: @Tofel what if there is an image for apple silicon? Shouldn't it be downloaded first? I wonder if you could use `platform := runtime.GOOS + "/" + runtime.GOARCH` to determine platform and first try to download native image. And, if it does not exist, fallback to any image.

lukaszcl: ah, I guess `NaiveRetrier` tries to download the native image? If yes, maybe but it first on the list in line [144](https://github.com/smartcontractkit/chainlink-testing-framework/pull/817/files#diff-3b425d7dadd18dc6d2fadc2bc169afae1f7bbc8f757a66d2837e9e78fb792c90R144)

Tofel: but it tries to download image for your current platform by default, and then this retry shouldn't even be triggered, no?

lukaszcl: ok, I thought we first use `Linux` `retriers = append(retriers, LinuxPlatoformImageRetrier, NaiveRetrier)`. But, these are retries and the first attempt is current platform (`ct, err = tc.GenericContainer(testcontext.Get(nil), req)`?

skudasov (approved):

docker/docker.go
Tofel: so that's how it works:
* first attempt is just whatever you passed in your container request
* if that fails, we go for retries
* first retry is with linux image platform (maybe here I could also check if the platform isn't already set to it, and if it is, skip that retry)
* then we try the naive retry, which is what we had before, which is just trying to start again without modifying anything

lukaszcl: great, thanks!

lukaszcl (approved):

PR #830, opened 2024-02-05
Enable Gnosis chain

davidcauchi:
Enable Gnosis chain


networks/known_networks.go
```diff
@@ -689,6 +689,34 @@ var (
...
+		MinimumConfirmations:      1,
+		GasEstimationBuffer:       1000,
+		FinalityTag:               true,
+		DefaultGasLimit:           6000000,
+	}
+
+	GnosisMainnet blockchain.EVMNetwork = blockchain.EVMNetwork{
+		Name:                      "Gnosis Mainnet",
+		SupportsEIP1559:           true,
+		ClientImplementation:      blockchain.GnosisClientImplementation,
```
AnieeG, line 709: Why not `EthereumClientImplementation` ?  Not seeing Gnosis specific implementation here.

davidcauchi: This has it's specific implementation in "blockchain/gnosis.go", also included in this PR. When checking with @kalverra this was the preferred approach rather than using `EthereumClientImplementation` directly when enabling a new chain.

AnieeG: Based on this PR `blockchain/gnosis.go `does not override/implement any method but just creates a wrapper for EthereumClient. What am I missing here? If you had used `EthereumClientImplementation`, would it not also serve the same purpose with less loc to maintain?

AnieeG (approved):

smickovskid (approved):

PR #835, opened 2024-02-15
TT-916 Add _secret suffix to secrets in toml

lukaszcl:
Mercury keeps secrets directly in TOML configs and uses a [masking script](https://github.com/smartcontractkit/mercury-pipeline/blob/main/.github/scripts/mask_toml_secrets.py#L0-L1) to hide them on Github CI.

In order to reuse TOML object from CTF like GrafanaConfig we need to add _secret suffix to all keys that should be masked on Github.

- [x] Create draft PRs for core, CCIP and Mercury to use new CTF lib and _secret keys. Only these repos are affected ([search query](https://github.com/search?q=org%3Asmartcontractkit++basic_auth%3D&type=code))
- [x] Check if Core works https://github.com/smartcontractkit/chainlink/pull/12043 confirmed with @Tofel 
- [x] Check if CCIP works https://github.com/smartcontractkit/ccip/pull/524 confirmed with @AnieeG 
- [x] Check if Mercury works https://github.com/smartcontractkit/mercury-pipeline/pull/732 @lukaszcl

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,8 +196,8 @@ When using `in-memory` or `file` target no other configuration variables are req
 [Logging.Loki]
 tenant_id="promtail"
 url="https://change.me"
-basic_auth="my-secret-auth"
-bearer_token="bearer-token"
+basic_auth_secret="my-secret-auth"
+bearer_token_secret="bearer-token"
 ```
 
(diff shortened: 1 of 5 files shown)


Tofel (approved):
LGTM

PR #838, opened 2024-02-23
Add Grafana HTTP API client

lukaszcl:


AnieeG (commented):
how about adding some tests?

grafana/client.go
```diff
@@ -0,0 +1,292 @@
...
+	Annotations  map[string]string `json:"annotations"`
+}
+
+func (p ProvisionedAlertRule) String() string {
+	var sb strings.Builder
+	sb.WriteString(fmt.Sprintf("ID: %d\n", p.ID))
+	sb.WriteString(fmt.Sprintf("UID: %s\n", p.UID))
+	sb.WriteString(fmt.Sprintf("FolderUID: %s\n", p.FolderUID))
+	sb.WriteString(fmt.Sprintf("Title: %s\n", p.Title))
+	// sb.WriteString("Data: [\n")
```
skudasov, line 232: Should we remove this?

grafana/client.go
```diff
@@ -0,0 +1,292 @@
...
+}
+
+// ruler API is deprecated https://github.com/grafana/grafana/issues/74434
+func (g *Client) GetAlertsForDashboard(dashboardUID string) (map[string][]interface{}, *resty.Response, error) {
+	var result map[string][]interface{}
+	r, err := g.resty.R().SetResult(&result).Get("/api/ruler/grafana/api/v1/rules?dashboard_uid=" + url.QueryEscape(dashboardUID))
+	return result, r, err
+}
+
+type AlertQuery interface{}
```
skudasov, line 211: Do we plan to add more on top of it? Empty type.

skudasov (approved):

grafana/client.go
lukaszcl: @skudasov I removed this type for now

skudasov (approved):

lukaszcl:
> how about adding some tests?

@AnieeG this PR contains API client for external server (Grafana). We discussed it with Sergey and we do not see any tests that would be useful for it at this point.

PR #840, opened 2024-02-27
[TT-1013] CLI tool to cleanup go test logs and make errors/panics more prominent

tateexon:
Adds tool to clean up go tests logs.
- cleans up testing.T.Log prefix
- adds color to errors and panics to make them more prominent
- adds parallel test formatting like gotestfmt but with several bug fixes
- can show only errors or both errors and passed tests
- for single package runs it can output tests as they finish to see results quicker
- Release script provided to push tag but release workflow will be add as part of [TT-1015], uses package.json to specify version and will use change sets to help promote good release documentation of the tools.

[TT-1015]: https://smartcontract-it.atlassian.net/browse/TT-1015?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tools/gotestloghelper/gotestevent/gotestevent_test.go
```diff
@@ -0,0 +1,349 @@
...
+
+	// Read the captured output from the channel
+	out := <-sr.outC
+
+	sr.isClosed = true
+
+	return out
+}
+
+// do not use with parallel tests
```
kalverra, line 66: When should we use it?

kalverra (approved):

tools/gotestloghelper/gotestevent/gotestevent_test.go
tateexon: Just when I thought I had cleared out all the unused and commented out code there is a massive block of it here. Good catch

kalverra (approved):

PR #842, opened 2024-02-27
[TT-941] Add better docker image version mirroring in ecr

tateexon:


.github/workflows/update-internal-mirrors.yaml
```diff
@@ -2,15 +2,92 @@ name: Update Mirrors
...
+          - name: nethermind/nethermind
+            expression: '^[0-9]+\.[0-9]+\.[0-9]+$'
+          # disabled until gcloud auth is added
+          # - name: gcr.io/prysmaticlabs/prysm/beacon-chain
+          #   expression: '^v[0-9]+\.[0-9]+\.[0-9]+$'
+          # - name: gcr.io/prysmaticlabs/prysm/validator
+          #   expression: '^v[0-9]+\.[0-9]+\.[0-9]+$'
+          - name: tofelb/ethereum-genesis-generator
+            expression: '^[0-9]+\.[0-9]+\.[0-9]+(\-slots\-per\-epoch)?'
+          # This one only has latest tag, probably only want to update it when we know for sure it's a new version we want
```
Tofel, line 51: 👍

README.md
```diff
@@ -330,4 +364,16 @@ Example:
...
 
 # TOML Config
-Basic and universal building blocks for TOML-based config are provided by `config` package. For more information do read [this](./config/README.md).
\ No newline at end of file
+
+Basic and universal building blocks for TOML-based config are provided by `config` package. For more information do read [this](./config/README.md).
+
+# ECR Mirror
+
+An ecr mirror can be used to push images used often in order to bypass rate limit issues from dockerhub. The list of image mirrors can be found in the [matrix here](./.github/workflows/update-internal-mirrors.yaml). This currently works with images with version numbers in dockerhub. Support for gcr is coming in the future. The images must also have a version number so putting `latest` will not work. We have a separate list for one offs we want that can be added to [here](./scripts/mirror.json) that does work with gcr and latest images. Note however for `latest` it will only pull it once and will not update it in our mirror if the latest on the public repository has changed, in this case it is preferable to update it manually when you know that you need the new latest and the update will not break your tests.
```
Tofel, line 372: could you add information on how to update manually?

.github/workflows/update-internal-mirrors.yaml
```diff
@@ -2,15 +2,92 @@ name: Update Mirrors
...
   update-mirrors:
+    runs-on: ubuntu-latest
+    environment: integration
+    strategy:
+      fail-fast: false
+      matrix:
+        mirror:
+          # note library just means it is part of the official library
+          - name: library/postgres
+            expression: '^[0-9]+\.[0-9]+$'
```
Tofel, line 23: so that means it will download all major and minor versions?

docker/test_env/non_dev_besu.go
```diff
@@ -28,7 +28,7 @@ import (
 )
 
 const (
-	BESU_IMAGE = "hyperledger/besu"
+	defaultNonDevBesuImage = "hyperledger/besu:24.1"
```
Tofel, line 31: I wonder if we are using this nondev Besu anywhere... as it might be totally broken. If I remember correctly the code was adjusted to PoW Besu and I never updated it. I think we should be using older Besu image here.

Tofel (approved):
nice one!

.github/workflows/update-internal-mirrors.yaml
tateexon: The script only pulls the latest 50 releases so in reality it only finds the latest release for each lts version:
16.2
15.6
14.11
13.14
12.18
Example from when I was testing the action: https://github.com/smartcontractkit/chainlink-testing-framework/actions/runs/8085942573/job/22094567790#step:5:40

PR #846, opened 2024-02-29
fix network policies for anvil and non-root-geth

Tofel:


charts/foundry/values.yaml
```diff
@@ -60,19 +60,30 @@ service:
...
+                  number: 8545
+
+networkPolicies:
+  enabled: true
+
+# Configure the default network policy.
+networkPolicyDefault:
+  ingress:
+    allowCustomCidrs: true
+    customCidrs: '10.13.0.0/19,10.13.32.0/19,10.13.64.0/19,10.13.96.0/19'
```
skudasov, line 86: Is it secure? Maybe comment on what these CIDRs belong to?

skudasov (approved):

chainchad (approved):

PR #859, opened 2024-03-07
[TT-950] Support for more eth1 clients

Tofel:
Tested with these EVM node ranges:
* Geth: 1.9.1 - 1.13.14
* Erigon: 2.40.0 - 2.58.1
* Nethermind: 1.16.0 -  1.25.4
* Besu: 22.1 - 24.1.2

Plus a bunch of other changes to private Ethereum network builder:
* consensus layer can be skipped now
* execution layer can be skipped if you're passing a custom docker image for it

Simplified version:
```
builder := NewEthereumNetworkBuilder()
_, err := builder.
  WithCustomDockerImages(map[ContainerType]string{
	  ContainerType_ExecutionLayer: fmt.Sprintf("hyperledger/besu:%s", AUTOMATIC_STABLE_LATEST_TAG)}).
Build()
```

Or with TOML:
```
[PrivateEthereumNetwork.CustomDockerImages]
execution_layer="hyperledger/besu:24.2.0-RC2"
```

Added two custom tags to our custom docker image handling:
* `latest_available`
* `latest_stable`

If you use them as shown in the example above we will fetch list of releases from Github and then either select latest one (even if it's a draft or pre-release) or latest stable.

Other changes:
* simplified and more universal structure for eth1 and eth2 clients
* Eth1 Geth always run with PoA consensus
* Eth1 Nethermind always runs with PoA consensus
* Eth1 Besu always runs with PoA consensus
* Eth1 Erigon always runs with PoW consensus (simulated `--fakepow`)
* `consensus_type` is deprecated and replaced with `ethereum_version` with two possible values: `eth1` and `eth2` (as these better capture the idea); backward compatibility is maintained.
* evmclient will fail if main wallet has 0 balance when running on simulated network (as it would fail anyway during funding attempt)
* `EVMClient` now has `GetEthClient()` method that return underlaying `ethclient.Client` (from `go-ethereum`)
* Dedicated method for restarting Postgres DB (used by some VRF tests) that reuses old containers (allows to avoid random errors in the CI)
* split docker-related e2e tests into 3 groups (defined by build tags) that run in 3 separate CI jobs
* better errors for starting eth1/eth2 containers

Tested in this [core PR](https://github.com/smartcontractkit/chainlink/pull/12332) with this [new pipeline](https://github.com/smartcontractkit/chainlink/actions/workflows/evm-version-compatibility-tests.yml) ✅

Tested in [this run](https://github.com/smartcontractkit/chainlink/actions/runs/8251104507) that bumping `go-ethereum` would have failed, if tested with EVM node in `v1.10.0`. ✅

CCIP was tested with this branch [here](https://github.com/smartcontractkit/ccip/actions/runs/8284832278/job/22671528447?pr=612) ✅


docker/test_env/eth2_common_test.go
```diff
@@ -1,3 +1,6 @@
+//go:build eth_env_tests
+// +build eth_env_tests
```
tateexon, line 2: I think we only need one of these. The top one is for go 1.17 and above while the lower one is for 1.16 and below. We are well beyond 1.17 at this point and I don't think we want to go back and support older versions that far back.

Tofel: for some reason some plugin in my VCS adds `go:build` on save even if I remove it... can we keep it? I think no harm done?

skudasov (changes requested):
LGTM 🚀 Let's figure out if we need build tags, fix some nitpicks and merge.

.github/workflows/lint.yaml
```diff
@@ -35,7 +35,7 @@ jobs:
         uses: golangci/golangci-lint-action@3a919529898de77ec3da873e3063ca4b10e7f5cc # v3.7.0
         with:
           version: v${{ needs.tools.outputs.golangci-lint-version }}
-          args: --out-format checkstyle:golangci-lint-report.xml
+          args: --out-format checkstyle:golangci-lint-report.xml --build-tags=other_docker_tests,besu_tests,geth_tests,erigon_tests,nethermind_tests,eth_env_tests
```
kalverra, line 38: Similar to other comments we've mentioned, I've had many a bad experience with build tags and tests, and would rather we only use them if absolutely necessary.

docker/test_env/eth2_common_test.go
kalverra: It's for compatibility reasons, no harm done.

README.md
```diff
@@ -133,19 +133,40 @@ We have extended support for execution layer clients in simulated networks. Foll
...
+
+```go
+builder := NewEthereumNetworkBuilder()
+cfg, err: = builder.
+    WithEthereumVersion(EthereumVersion_Eth2).
+    WithExecutionLayer(ExecutionLayer_Geth).
+    Build()
+```
+
+Since we support both `eth1` (aka pre-Merge) and `eth2` (aka post-Merge) client versions, you need to specify which one you want to use. You can do that by calling `WithEthereumVersion` method. There's no default provided. The only execption is when you use custom docker images (instead of default ones), because then we can determine which version it is based on the image version.
```
kalverra, line 146: Can we mention a scenario, or just generally why I would want to use one or the other?

kalverra (changes requested):

README.md
Tofel: That is a very good question, why would you want to use this ethereum version or that one? I think the only reason is block speed.

docker/test_env/besu_test.go
```diff
@@ -15,12 +15,36 @@ import (
 	"github.com/smartcontractkit/chainlink-testing-framework/utils/testcontext"
 )
 
+func TestEth1WithBesu(t *testing.T) {
```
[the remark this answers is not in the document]
Tofel: added a ticket

skudasov (approved):

kalverra (approved):

PR #866, opened 2024-03-14
Revert WaspAutoBuildConfig

lukaszcl:
This is still needed in core. At least until core starts to use new K8s Test Runner

### DIFF
--- a/config/wasp_autobuild.go
+++ b/config/wasp_autobuild.go
@@ -1,0 +1,39 @@
+package config
+
+import (
+	"errors"
+)
+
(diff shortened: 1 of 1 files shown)


Tofel (approved):

config/wasp_autobuild.go
```diff
@@ -0,0 +1,39 @@
+package config
+
+import (
+	"errors"
+)
+
+type WaspAutoBuildConfig struct {
```
AnieeG, line 7: please add a test to avoid removal in future

PR #870, opened 2024-03-19
two small upgrades

Tofel:
* fix network naming
* set chain id in EVMNetwork to what was passed in toml config


tateexon (approved):
I just ask that you make a ticket to fix the rolloutrestartupdate tests since you are skipping them now.

PR #871, opened 2024-03-20
remove nondev geth/besu code

Tofel:
This PR removed non-dev geth/besu version from docker test environment.

Core PR that run on these changes is available [here](https://github.com/smartcontractkit/chainlink/pull/12506).
CCIP PR run without non-dev geth/besu is available [here](https://github.com/smartcontractkit/ccip/pull/626)


docker/test_env/non_dev_geth.go
```diff
@@ -1,426 +0,0 @@
...
-	"github.com/smartcontractkit/chainlink-testing-framework/utils/templates"
-	"github.com/smartcontractkit/chainlink-testing-framework/utils/testcontext"
-)
-
-const (
-	TX_NON_DEV_GETH_WS_PORT = "8546"
-	BOOTNODE_PORT           = "30301"
-
-	defaultNonDevGethImage         = "ethereum/client-go:v1.12.0"
-	defaultNonDevGethAllToolsImage = "ethereum/client-go:alltools-v1.10.25"
```
tateexon, line 36: Is anything else using this image? If not we should remove it from the mirror workflow.

tateexon (approved):

docker/test_env/non_dev_geth.go
Tofel: afaik nobody else is using it, we can remove them

PR #884, opened 2024-03-27
BCI-2646 - Adding zkSync tests 

smickovskid:
- BCI-2646 - Adding zkSync tests


Tofel:
closing due to staleness; we have also decided to use another approach, when deploying to zksync

PR #890, opened 2024-03-27
[TT-1015] Add release workflow for tools

tateexon:


tateexon:
Closing since this doesn't work with the k8s-test-runner, it has issues and may just need a proper build for a release. Will rework for just the simple go cli tools

PR #904, opened 2024-04-05
Add grafana-agent pod monitor for scraping pg-exporter metrics

emate:
Add grafana-agent pod monitor to scrape metrics from prometheus-postgres-exporter

### DIFF
--- a/charts/chainlink/templates/pg-deployment.yaml
+++ b/charts/chainlink/templates/pg-deployment.yaml
@@ -38,6 +38,9 @@ spec:
         {{- range $key, $value := $.Values.db.labels }}
         {{ $key }}: {{ $value | quote }}
         {{- end }}
+        {{- if $cfg.db.enablePrometheusPostgresExporter }}
+        prometheus-postgres-exporter: true
+        {{- end }}
       annotations:
         prometheus.io/scrape: 'true'
     spec:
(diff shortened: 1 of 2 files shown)


charts/chainlink/templates/pg-deployment.yaml
```diff
@@ -38,6 +38,9 @@ spec:
         {{- range $key, $value := $.Values.db.labels }}
         {{ $key }}: {{ $value | quote }}
         {{- end }}
+        {{- if $cfg.db.enablePrometheusPostgresExporter }}
```
AnieeG, line 41:
```suggestion
        {{- if $.Values.db.enablePrometheusPostgresExporter }}
```

charts/chainlink/templates/pod-monitor.yaml
```diff
@@ -10,4 +10,19 @@ spec:
       release: {{ .Release.Name }}
   podMetricsEndpoints:
     - port: access
-{{- end }}
\ No newline at end of file
+---
+{{- if .Values.db.enablePrometheusPostgresExporter }}
```
AnieeG, line 14: does it work with this or `$` is needed?

AnieeG (approved):

charts/chainlink/templates/pod-monitor.yaml
emate: Yes, it will work fine, as it's not a nested block.

PR #905, opened 2024-04-06
Fix prometheus-postgres-exporter manifest

emate:
Fix broken prometheus-postgres-exporter manifest


charts/chainlink/templates/pod-monitor.yaml
```diff
@@ -2,6 +2,7 @@
 apiVersion: monitoring.coreos.com/v1
 kind: PodMonitor
 metadata:
+  name: "{{ .Release.Name }}-pod-monitor"
```
AnieeG, line 5: This works without the quotes I think

AnieeG (approved):

PR #907, opened 2024-04-09
Fix lint in k8s test runner

lukaszcl:


k8s-test-runner/runner/runner.go
```diff
@@ -2,6 +2,7 @@ package runner
 
 import (
 	"context"
+	// "embed"
```
skudasov, line 5: Should we remove it?

skudasov (approved):

k8s-test-runner/runner/runner.go
lukaszcl: @skudasov otherwise it complains about `_ "embed"`

PR #912, opened 2024-04-12
Removes the `ManifestUpdate` Double Negative

kalverra:
---

## Why
The changes improve clarity and consistency in environment variable naming and usage across Kubernetes-related configurations and tests. Specifically, they switch from a negative to a positive naming convention for manifest updates, making it easier to understand and use.

## What
- **k8s/config/overrides.go**
  - Renamed environment variable from `NO_MANIFEST_UPDATE` to `MANIFEST_UPDATE`. This change affects how manifest updates are controlled, aligning the variable name with its purpose more clearly.
- **k8s/e2e/common/test_common.go**
  - Updated references to the renamed environment variable `MANIFEST_UPDATE` in test configurations and logic. This ensures that tests correctly interpret the flag for manifest updating.
- **k8s/e2e/remote-runner/remote_runner_envs_test.go**
  - Adjusted test configurations to use the new `MANIFEST_UPDATE` environment variable, ensuring that remote runner tests adhere to the updated manifest update control logic.
- **k8s/environment/environment.go**
  - Reflected the environment variable change in the environment configuration, switching from `NoManifestUpdate` to `ManifestUpdate`. This modification impacts how environments handle manifest updates during their setup and execution.
- **k8s/environment/runner.go**
  - Updated the job environment variables to include the renamed `MANIFEST_UPDATE`, which is propagated to remote runners, affecting how they manage manifest updates.


k8s/e2e/common/test_common.go
```diff
@@ -126,13 +126,13 @@ func TestConnectWithoutManifest(t *testing.T) {
...
 		require.NotEmpty(t, os.Getenv(config.EnvVarNamespace))
-		noManifestUpdate, err := strconv.ParseBool(os.Getenv(config.EnvVarNoManifestUpdate))
+		noManifestUpdate, err := strconv.ParseBool(os.Getenv(config.EnvVarManifestUpdate))
 		require.NoError(t, err, "Failed to parse the no manifest update env var")
 		require.True(t, noManifestUpdate)
 	}
 
 	// Now run an environment without a manifest like a normal test
-	testEnvConfig.NoManifestUpdate = true
+	testEnvConfig.ManifestUpdate = true
```
AnieeG, line 135: if NoManifestUpdate was true, should this not be false?

kalverra:
> Would it make sense to just rename it to SKIP_MANIFEST_UPDATE? My reasoning is that we only even provide this when we want to skip it.

Ya @tateexon I like this idea better, even making sure all the reversing is making my head hurt.

.pre-commit-config.yaml
```diff
@@ -19,10 +19,8 @@ repos:
     hooks:
       - name: Prettier
         id: prettier
-  - repo: local
+  - repo: https://github.com/golangci/golangci-lint
```
tateexon, line 22: I just recently changed this so it would use the version we either have installed via asdf or nix and we don't have to keep yet another version up to date here.

PR #913, opened 2024-04-12
Adds Pre Commit Hook

kalverra:
Also adds explanation for how to use it + dependency setup

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -1,0 +1,86 @@
+#!/bin/bash
+
+echo "Running pre-commit checks..."
+
+has_errors=0
+
+# Run go mod tidy
+echo -n "Checking go mod tidy..."
+
+go mod tidy
+
+# Check if any changes were made
+if [[ $(git diff --name-only) ]]; then
+    echo -e "\rChecking go mod tidy... ❌"
+    echo "Changes made by 'go mod tidy'. Please commit the changes."
+    has_errors=1
+else
+    echo -e "\rChecking go mod tidy... ✅"
+fi
+
+# Define the regex pattern for Ethereum keys
+ethereum_key_pattern="(0x[a-fA-F0-9]{40})"
+# Define the regex pattern for HTTP and WS URLs
+url_pattern="(http|ws)(s)?:\/\/[^ \t\n\r]+"
+# Check for staged files
+staged_files=$(git diff --cached --name-only)
+
+echo -n "Checking for Ethereum keys, RPC URLs, and DEBUG: in staged files..."
+# Iterate through staged files
+for file in $staged_files; do
+    # Check if the file is tracked
+    if git ls-files --error-unmatch "$file" &>/dev/null; then
+        # Check if the file contains an Ethereum key
+        if grep -q "$ethereum_key_pattern" "$file"; then
+            has_errors=1
+            echo -e "\rChecking for Ethereum keys, RPC URLs, and DEBUG: in staged files... ❌"
+            echo "ERROR: Ethereum key detected in file: $file"
+            grep -n "$ethereum_key_pattern" "$file"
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -16,6 +16,20 @@ The Chainlink Testing Framework is a blockchain development framework written in
...
+
+```sh
+make install
+```
+
+We also recommend using our handy githooks.
+
+```sh
+git config core.hooksPath .githooks
+```
```
tateexon, line 31: Do you want to instead just replace the `pre-commit install` that exists in the nix file and the `make install`, otherwise they will get a different pattern. If they don't run this command then they get both the pre-commit and pre-push running on commit

tateexon (changes requested):

README.md
kalverra: Yes, I'm unfamiliar with nix though. I'm assuming the answer is the `pre-commit-config.yaml`?

tateexon: Just change in it nix here: https://github.com/smartcontractkit/chainlink-testing-framework/blob/main/shell.nix#L60

.pre-commit-config.yaml
```diff
@@ -21,8 +22,14 @@ repos:
...
   - repo: local
     hooks:
-      - id: golangci-lint
-        name: golangci-lint
-        entry: golangci-lint run
-        language: system
-        types: [go]
+      - id: custom-pre-commit
+        name: Custom pre-commit hook
+        entry: ./githooks/pre-commit
```
tateexon, line 27:
```suggestion
        entry: ./.githooks/pre-commit
```

.pre-commit-config.yaml
```diff
@@ -21,8 +22,14 @@ repos:
...
+        language: script
+        stages: [commit]
+      - id: custom-pre-push
+        name: Custom pre-push hook
+        entry: ./githooks/pre-push
```
tateexon, line 32:
```suggestion
        entry: ./.githooks/pre-push
```

.pre-commit-config.yaml
```diff
@@ -21,8 +22,14 @@ repos:
...
+        language: script
+        stages: [push]
```
tateexon, line 34:
```suggestion
        stages: [pre-push]
```
As per the documentation: https://pre-commit.com/#confining-hooks-to-run-at-certain-stages

.pre-commit-config.yaml
```diff
@@ -21,8 +22,14 @@ repos:
...
-      - id: golangci-lint
-        name: golangci-lint
-        entry: golangci-lint run
-        language: system
-        types: [go]
+      - id: custom-pre-commit
+        name: Custom pre-commit hook
+        entry: ./githooks/pre-commit
+        language: script
+        stages: [commit]
```
tateexon, line 29:
```suggestion
        stages: [pre-commit]
```
As per the documentation: https://pre-commit.com/#confining-hooks-to-run-at-certain-stages

tateexon (changes requested):

.githooks/pre-commit
```diff
@@ -0,0 +1,150 @@
...
+else
+    printf "Checking 'go mod tidy'... ✅\n"
+fi
+
+printf "Checking go.mod for local replacements...\r"
+
+if grep -q "replace .* => /" "./go.mod"; then
+    has_errors=1
+    printf "Checking go.mod for local replacements... ❌\n"
+    grep -n "replace .* => /" "./go.mod"
```
tateexon, line 32: This only checks for local replaces that use absolute paths. Is this common? I may just do things weird but my local replaces are usually using relative paths so would start with ../ and wouldn't get caught by this, but if we did that here then we would block things like sub projects that use the main project as a relative local replace.

PR #922, opened 2024-04-19
set EIP-1559 to not supported in network config for Geth v1.10.x and …

Tofel:
…lower

---

## Why
The changes ensure compatibility and accurate configuration for different versions of Geth within test environments, specifically addressing network configuration based on the Geth version being used.

## What
- **docker/test_env/geth_base.go**
  - Added condition to set `networkConfig.SupportsEIP1559 = false` if `comparableVersion < 110`. This adjustment ensures that the test environment correctly reflects the capabilities of Geth versions prior to 1.10.x, where `eth_maxPriorityFeePerGas` is not implemented.


Tofel:
repo cleanup: no longer needed

PR #923, opened 2024-04-23
Add show_coverage_report to logging conf

lukaszcl:
Required for https://github.com/smartcontractkit/chainlink/pull/12835

---

## Why
The update introduces a new configuration option to display HTML coverage reports, enhancing the logging configuration's flexibility and utility for monitoring and analysis purposes.

## What
- **config/logging.go**
  - Added `ShowHTMLCoverageReport *bool 'toml:"show_html_coverage_report"'` to `LoggingConfig` structure. This change allows users to configure the visibility of HTML coverage reports.


config/logging.go
```diff
@@ -9,11 +9,12 @@ import (
 )
 
 type LoggingConfig struct {
-	TestLogCollect *bool            `toml:"test_log_collect"`
-	RunId          *string          `toml:"run_id"`
-	Loki           *LokiConfig      `toml:"Loki"`
-	Grafana        *GrafanaConfig   `toml:"Grafana"`
-	LogStream      *LogStreamConfig `toml:"LogStream"`
+	TestLogCollect         *bool            `toml:"test_log_collect"`
+	ShowHTMLCoverageReport *bool            `toml:"show_html_coverage_report"` // Show reports with go coverage data
```
AnieeG, line 13: can we make it optional by providing `omitempty`?

AnieeG: Yes I would prefer it to be `omitempty` for all

lukaszcl: @AnieeG `toml:",omitempty"`?

AnieeG: Right. `toml:",omitempty"`

Tofel: you mean `toml:"<CURRENT_NAME>,omitempty"`, right? :-)

lukaszcl: @Tofel I think `toml:",omitempty"`. I see this is how core sets it https://github.com/smartcontractkit/chainlink/blob/develop/core/config/toml/types.go#L43. @Tofel @AnieeG  pls confirm

AnieeG: @Tofel has been following a convention of differentiating the names for tablenames and keyname under the table.  Like 

> [Network] is table name and `selected_networks` is a key name that's under that table.

If we decide to avoid doing that, skipping explicit names makes sense.

lukaszcl: @kalverra thanks for explaining it. I added `omitempty` in next commit

AnieeG (approved):

PR #924, opened 2024-04-23
Fixes to foundry connections and configs

AnieeG:
---
Anvil can be run without forking. Renamed network's `ForkConfig` to `AnvilConfig`. Made necessary changes to make fork config not mandatory while using Anvil.

## Why
The changes improve the function signatures in `foundry.go` by switching from value to pointer receivers for the `Props` parameter. This enhances memory efficiency and consistency in how the `Props` struct is handled across functions.

## What
- **k8s/pkg/helm/foundry/foundry.go**
  - Modified the `New` function signature to accept a pointer to `Props` instead of a value. This change optimizes memory usage by passing a reference.
  - Altered the `NewVersioned` function signature to also accept a pointer to `Props` for the same reasons of efficiency and consistency with the `New` function adjustments.

### DIFF
--- a/config/network.go
+++ b/config/network.go
@@ -18,13 +18,13 @@ const (
 )
 
 type ForkConfig struct {
-	URL              string `toml:"url"`                          // URL is the URL of the node to fork from. Refer to https://book.getfoundry.sh/reference/anvil/#options
-	BlockNumber      int64  `toml:"block_number"`                 // BlockNumber is the block number to fork from. Refer to https://book.getfoundry.sh/reference/anvil/#options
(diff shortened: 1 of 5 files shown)


config/network.go
```diff
@@ -149,14 +149,8 @@ func (n *NetworkConfig) Validate() error {
 				return fmt.Errorf("chain ID for %s network must be set", name)
 			}
 		}
-		if n.ForkConfigs != nil {
-			if _, ok := n.ForkConfigs[network]; ok {
```
kalverra, line 153: Why remove these checks?

AnieeG: While using Anvil you may choose not to fork a network, and start network from empty state like geth. Hence these checks are not required

kalverra (approved):

PR #928, opened 2024-04-25
Change blocktime to float

AnieeG:
---

## Why
The change updates the `BlockTime` field type from `int64` to `float64` in the AnvilConfig struct to allow more granular block time configurations.

## What
- **config/network.go**
  - Changed the type of `BlockTime` from `int64` to `float64` in the `AnvilConfig` struct. This change allows for more precise block time configurations, accommodating the need for sub-second block times.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ lint:
 
 .PHONY: tidy
 tidy:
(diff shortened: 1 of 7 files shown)


Tofel (approved):

AnieeG:
Anvil is not accepting sub second blocktime.closing this

PR #945, opened 2024-05-08
Foundry persistence - dump foundry state in an interval of 10s

AnieeG:


charts/foundry/templates/deployment.yaml
```diff
@@ -104,7 +104,10 @@ spec:
             - "--balance"
             - {{ .Values.anvil.accountBalance | quote }}
             {{- end }}
-            - "--prune-history"
+            - "--state"
+            - "/data/dump.json"
+            - "--state-interval"
+            - "10"
```
scheibinger, line 110: If the data dump happens on the intervals and is not in the a realtime, maybe it would be better to push this to S3. How large is the data dump?

Does `10` means, dump data every 10 seconds?

AnieeG: yes right ! it dumps every 10s to the file location specified in `--state` args and during start up it loads from that file if the file exists

charts/foundry/values.yaml
```diff
@@ -54,10 +54,6 @@ securityContext:
     drop:
     - ALL
   readOnlyRootFilesystem: false
-  runAsNonRoot: true
```
AnieeG, line 57: @skudasov I had to remove this, otherwise Anvil is getting into permission denied error while creating dump state file

charts/foundry/templates/deployment.yaml
```diff
@@ -104,7 +104,12 @@ spec:
             - "--balance"
             - {{ .Values.anvil.accountBalance | quote }}
             {{- end }}
-            - "--prune-history"
+            - "--state"
+            - "/state/dump.json"
+            - "--state-interval"
+            - "60"
+            - "--transaction-block-keeper"
```
skudasov, line 111: Not sure it's even documented there https://github.com/foundry-rs/foundry/issues/5055 maybe we should put it in `Values.anvil.transaction-block-keeper` since we might tune it in the future?

skudasov (changes requested):

skudasov (approved):

charts/foundry/values.yaml
scheibinger: AFAIR, if you remove this, your deployment would be blocked by a OPA gatekeeper policy engine. Were you able to deploy it?

charts/foundry/values.yaml
```diff
@@ -54,10 +56,6 @@ securityContext:
...
-  # foundry user/group
-  runAsUser: 1000
-  runAsGroup: 1000
```
scheibinger, line 59: try adding `fsGroup: 1000` here I think that should help to make the volume writable from non root user

scheibinger: you can find more info here: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#configure-volume-permission-and-ownership-change-policy-for-pods

PR #957, opened 2024-05-16
[TT-1164] gotestloghelper no tests printed out on segfault

tateexon:
Also now highlights panics and seg faults as red.
This also covers [TT-1101]

[TT-1101]: https://smartcontract-it.atlassian.net/browse/TT-1101?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

## Why
The changes enhance the gotestloghelper tool by improving panic error handling, refining test output, and upgrading to Go 1.21.9 for compatibility and performance improvements. These modifications aim to provide clearer, more informative test logs and ensure the tool's stability with the latest Go version.

## What
- **tools/gotestloghelper/go.mod**
  - Updated Go version from 1.21.5 to 1.21.9. This change ensures compatibility with the latest Go features and improvements.
- **tools/gotestloghelper/gotestevent/gotestevent.go**
  - Modified panic constants and regular expressions for better panic detection and handling. Introduced `testPanicWithTest` and updated `testPanic` regex to cover scenarios including SIGSEGV signals.
  - Adjusted the `Print` method's default outcomeType from "pass" to "fail" to reflect a more cautious approach when determining a test's outcome.
  - Enhanced error handling and output in the `Print` method to better distinguish between different types of test failures.
  - Made `Print` method of `TestPackage` a pointer receiver, aligning with Go best practices for methods that modify the receiver's state.
  - Added logic to clear out printed test logs to save memory and prevent double printing, optimizing runtime efficiency.
- **tools/gotestloghelper/gotestevent/gotestevent_test.go**
  - Implemented additional test cases to verify the handling of panic scenarios, including segmentation faults and unexpected panic formats. This ensures the tool robustly captures and reports a wider range of error conditions in test execution.
  - Adjusted test assertions to ignore differences in whitespace, focusing on the structural accuracy of the output rather than formatting, which enhances test reliability.
- **tools/gotestloghelper/package.json**
  - Incremented the version number from 1.0.3 to 1.0.4 to mark the introduction of new features and improvements in the tool.


jmank88:
What is the purpose of detecting panic logs? This is brittle at best, and still limited only to panics.

When a panic occurs during testing, the whole suite fails. The same is true for fatal exits. In both cases, it is not possible in the general case to attribute them to a particular test. In fact, there may be no associated test at all. The entire package suite fails, and we must dump all std out logs from that package. IMHO we should drop specialized panic logic and replace it with general, robust handling of packages failures. Panics will naturally be presented normally, without any special treatment.

kalverra (approved):

tateexon:
@jmank88 
> What is the purpose of detecting panic logs? This is brittle at best, and still limited only to panics.
> 
> When a panic occurs during testing, the whole suite fails. The same is true for fatal exits. In both cases, it is not possible in the general case to attribute them to a particular test. In fact, there may be no associated test at all. The entire package suite fails, and we must dump all std out logs from that package. IMHO we should drop specialized panic logic and replace it with general, robust handling of packages failures. Panics will naturally be presented normally, without any special treatment.

We do handle the package failing outside of whether the panic regex and stuff fails: https://github.com/smartcontractkit/chainlink-testing-framework/blob/main/tools/gotestloghelper/gotestevent/gotestevent.go#L400-L406

The more brittle regex stuff is to try to make certain parts pop out a bit more by highlighting them in red/yelow/green or grabbing some useful information like panics and errors and showing them to the developer at the top so they don't have to dig through potentially a thousand lines of logs to get a quick idea of why it failed. I don't know a great way to not have these parts be brittle since they are subjective to a large extent on what we think is important and not necessarily something that is spelt out in the data. We do hope to plan out and use our AI hooks to replace a lot of these regex parts by having the AI grab what is useful and present it in a developer friendly way but that is probably more Q4 time frame.

jmank88:
> We do handle the package failing outside of whether the panic regex and stuff fails:

If this is true, then why don't we see all the logs? AFAICT this is broken

tateexon:
@jmank88 Are you saying you want ALL the logs or all the logs for the package that had the failure? We do the latter but not the former. This particular PR fixes the later for cases when we run in -singlepackage mode where we try to print the tests out as they finish but shouldn't affect how this was working for the core unit tests since that was not running in -singlepackage mode.

jmank88:
> @jmank88 Are you saying you want ALL the logs or all the logs for the package that had the failure? We do the latter but not the former. This particular PR fixes the later for cases when we run in -singlepackage mode where we try to print the tests out as they finish but shouldn't affect how this was working for the core unit tests since that was not running in -singlepackage mode.

We currently have package failures on normal core tests which do not dump any logs.

jmank88:
> We currently have package failures on normal core tests which do not dump any logs.

https://github.com/smartcontractkit/chainlink/actions/runs/8826060246/job/24232296151?pr=12959

PR #960, opened 2024-05-17
[TT-1179] make log scanner more robuts by creating a temporary snapshot

Tofel:
---

## Why
The changes introduce a more robust way to handle log reading during log processing by creating a temporary snapshot of the log file. This method prevents potential data corruption by locking the log file against new writes while the snapshot is created and used for reading.

## What
- `logstream/logprocessor.go`
  - Modified `ProcessContainerLogs` function to use a temporary snapshot for log processing instead of duplicating the file descriptor. This change ensures that the log processing does not interfere with concurrent log writes.
  - Added `createTemporarySnapshot` function that creates a temporary file snapshot of the log file. This function duplicates the file descriptor, creates a temporary file, copies the contents of the original log file to the temporary file, and verifies the integrity of the copied data.

### DIFF
--- a/logstream/logprocessor.go
+++ b/logstream/logprocessor.go
@@ -40,18 +40,17 @@ func (l *LogProcessor[ReturnType]) ProcessContainerLogs(containerName string, pr
 		return new(ReturnType), fmt.Errorf("no consumer found for container %s", containerName)
 	}
 
-	// Duplicate the file descriptor for independent reading, so we don't mess up writing the file by moving the cursor
-	fd, err := syscall.Dup(int(consumer.tempFile.Fd()))
+	// Create a temporary snapshot of the log file and temporarily lock accept mutex to prevent new logs from being written
+	// as that might corrupt the gob file due to saving of incomplete logs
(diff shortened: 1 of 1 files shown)


logstream/logprocessor.go
```diff
@@ -104,3 +103,51 @@ func GetRegexMatchingProcessor(logStream *LogStream, pattern string) (*LogProces
...
 	return logProcessor, processFn, nil
 }
+
+func createTemporarySnapshot(file *os.File) (*os.File, error) {
+	// Duplicate the file descriptor (so that when we work with the file, we don't affect the original file descriptor, especially cursor position)
+	fd, err := syscall.Dup(int(file.Fd()))
+	if err != nil {
+		return nil, err
+	}
+	readFile := os.NewFile(uintptr(fd), "name_doesnt_matter.txt")
```
kalverra, line 113: nit: This filename will be confusing to me if I catch it. Best to name it something like `snapshot.txt`

kalverra (changes requested):

kalverra (approved):

PR #965, opened 2024-05-23
[CCIP-2155] Enable Easier Wallet Creation

kalverra:
There are CCIP test scenarios that require contract calls to come from non-admin addresses.

---

## Why
The changes introduce a new functionality that allows the generation and funding of new Ethereum wallets directly through the EVMClient interface. This enhancement facilitates dynamic wallet management within blockchain applications, enabling the creation of wallets on-the-fly and optionally funding them as part of the wallet creation process.

## What
- **blockchain/blockchain.go**
  - Added a new method signature `NewWallet(funding *big.Int) (int, error)` to the `EVMClient` interface. This method aims to generate and optionally fund a new wallet, returning its index within the client's wallet list.
- **blockchain/ethereum.go**
  - Implemented the `NewWallet` method within `EthereumClient` and `EthereumMultinodeClient`. This implementation generates a new Ethereum wallet, optionally funds it if a non-nil funding amount is specified, and appends the new wallet to the client's wallet list. This method returns the index of the newly created wallet within the list or an error if the operation fails.


blockchain/ethereum.go
```diff
@@ -170,6 +170,29 @@ func (e *EthereumClient) GetWallets() []*EthereumWallet {
...
+	privateKeyHex := hex.EncodeToString(privateKeyBytes)
+	newWallet, err := NewEthereumWallet(privateKeyHex)
+	if err != nil {
+		return -1, err
+	}
+	if funding != nil {
+		err = e.Fund(newWallet.Address(), conversions.WeiToEther(funding), GasEstimations{})
+		if err != nil {
+			return -1, err
+		}
```
b-gopalswami, line 190: As there is no data, can do inline error check.
```suggestion
		if err = e.Fund(newWallet.Address(), conversions.WeiToEther(funding), GasEstimations{}); err != nil {
			return -1, err
		}
```

b-gopalswami (approved):
LGTM!

blockchain/ethereum.go
kalverra: Applied, Ty

tateexon (approved):

PR #1004, opened 2024-06-20
[TT-1278] exclude spaces or dashes when matching network names

Tofel:
Why? Now it's a bit painful to remember that if we use this as network:
```
selected_networks=["POLYGOIN_MANNET"]
```
then we need to define Seth network as:
```
[[Seth.Network]]
name="POLYGON MAINNET"
```

(notice space instead of `_`) 

That because currently we match by `EVMNetwork`'s `Name` field and not the key under which it's stored in known networks map.

---

## Why
The changes introduce enhancements in network configuration handling within the Seth module, specifically improving the flexibility and accuracy in matching network names between the EVM and Seth configurations. Additionally, the inclusion of a regular expression utility aims to normalize network name comparisons, ensuring more reliable and consistent configuration merging processes.

## What
- **`utils/seth/seth.go`**
  - Added `regexp` package import to utilize regular expression capabilities for network name normalization.
  - Modified `MergeSethAndEvmNetworkConfigs` function to utilize a new helper function `isSameNetwork` for network name comparison, replacing a case-insensitive string comparison. This change aims to improve the logic for matching network names by considering variations in naming conventions (e.g., differences in whitespace or hyphens).
  - Added `isSameNetwork` function which normalizes network names by converting spaces and hyphens to underscores before performing a case-insensitive comparison. This ensures a more accurate and flexible way of determining if a given EVM network and Seth network configuration refer to the same underlying network, accounting for minor variations in how network names might be presented or formatted.

### DIFF
--- a/utils/seth/seth.go
+++ b/utils/seth/seth.go
@@ -2,6 +2,7 @@ package seth
 
 import (
 	"fmt"
+	"regexp"
 	"strings"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


lukaszcl (approved):
@Tofel approved. Btw, I guess we can't have 1 place to store network names / constants that can be used by different go modules (CTF, Seth, integration-tests in Core, etc.)?

PR #1005, opened 2024-06-20
More Consistent Keys are Easier to Fund

kalverra:
---

## Why
The changes update private keys and introduce a new private key for certain simulated and non-simulated blockchain networks. These modifications ensure the networks have the correct credentials for authentication and maintain the integrity and security of transactions on these networks.

## What
- **networks/known_networks.go**
  - Updated private keys for `SimulatedEVMNonDev1`, `SimulatedBesuNonDev1`, and `SimulatedEVMNonDev` networks.
  - Added a new private key to `SimulatedEVMNonDev2`, `SimulatedBesuNonDev2`, and several other networks, enhancing their security configuration.

### DIFF
--- a/networks/known_networks.go
+++ b/networks/known_networks.go
@@ -56,7 +56,7 @@ var (
 		ChainID:              1337,
 		PrivateKeys: []string{
 			"ac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80",
-			"7c852118294e51e653712a81e05800f419141751be58f605c371e15141b007a6",
+			"59c6995e998f97a5a0044966f0945389dc9e86dae88c7a8412f4603b6b78690d",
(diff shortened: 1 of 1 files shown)


networks/known_networks.go
```diff
@@ -122,8 +122,8 @@ var (
 		SimulationType:       "besu",
 		ChainID:              2337,
 		PrivateKeys: []string{
+			"ac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80",
```
AnieeG, line 125: Hey let's make this as second key for chain id with 2337. The reason - if you use same private key for 1337 and 2337 it will create contracts on both chains with same address. It's not an issue but it becomes a little bit difficult during debugging.

AnieeG (approved):

PR #1039, opened 2024-08-06
Add ReadEnvVarSlice_String to read config values

lukaszcl:
---

## Why
The changes improve the environment variable handling capabilities of the application by introducing a new function for reading slices of strings based on a pattern and optimizing how grouped environment variables are read.

## What
- **config/env.go**  
  - Added `ReadEnvVarSlice_String` function: This new function allows the application to read and return slices of strings from environment variables that match a specified regex pattern. It iterates through all environment variables, matches keys against the pattern, and collects non-empty values.
  - Modified `ReadEnvVarGroupedMap` function: Previously, this function directly merged maps without checking if the `singleEnvPattern` was non-empty. Now, it conditionally initializes and fills `singleMap` based on `singleEnvPattern` before merging, ensuring that only relevant environment variables are considered.


config/env.go
```diff
@@ -54,8 +54,29 @@ func MustReadEnvVar_Boolean(name string) *bool {
 	return ptr.Ptr(value.(bool))
 }
 
+func ReadEnvVarSlice_String(pattern string) []string {
```
Tofel, line 57: any reason for breaking with convention and using `_` in method name?

Tofel: Maybe it should also have a suffix of `Must`? Since it doesn't return error and panics, when regex is incorrect? Or... maybe it should accept `re` instead of `string` and then there would be no problem?

config/env.go
```diff
@@ -54,8 +54,29 @@ func MustReadEnvVar_Boolean(name string) *bool {
...
+		}
+		key, value := pair[0], pair[1]
+		if re.MatchString(key) && value != "" {
+			values = append(values, value)
+		}
+	}
+	return values
+}
+
 func ReadEnvVarGroupedMap(singleEnvPattern, groupEnvPattern string) map[string][]string {
```
Tofel, line 74: add comments to both methods, ok?

lukaszcl: pushed commit with comments

config/env.go
lukaszcl: @Tofel no must since it returns empty slice if no vars found

gheorghestrimtu (approved):

PR #1082, opened 2024-08-21
Merge WASP library

skudasov:


.github/workflows/wasp-lint.yml
```diff
@@ -0,0 +1,27 @@
...
+  contents: read
+jobs:
+  golangci:
+    defaults:
+      run:
+        working-directory: wasp
+    name: lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
```
lukaszcl, line 14: @skudasov why not to use `paths` from github? https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore I think way the workflow will not be started at all

skudasov: We've decided to go with `dorny`'s action for now since GHA way of solving optional pipelines and required checks is not that good https://github.com/orgs/community/discussions/44490

gheorghestrimtu (approved):

PR #1092, opened 2024-08-26
[TT-1511] added ability to read ABIs from Gethwrappers

Tofel:
Background:
Because new deployer won't be using Seth's `DeployContract()` method we need a new way to read contract ABIs so that tracing can work. That new way will walk through specified directories looking for gethwrappers and load ABIs from them.

- [x] allow to read ABIs from geth wrappers
- [x] support multiple folders with wrappers
- [x] make it fault-tolerant, but fail on empty or invalid ABIs
- [x] add geth wrappers folder to `ClientBuilder`

Tests:
- [x] load both ABI and gethwrappers -> PASS
- [x] load only gethwrappers -> PASS
- [x] load gethwrappers from multiple directories -> PASS
- [x] load gethwrappers fro directory with wrappers and regular go files -> PASS
- [x] load gethwrapper with empty ABI -> FAIL
- [x] load gethwrapper with broken ABI -> FAIL
- [x] load gethwrappers from non-existent directory -> FAIL


lukaszcl (approved):

seth/contract_store.go
```diff
@@ -15,6 +20,7 @@ const (
 	ErrOpenABIFile = "failed to open ABI file"
 	ErrParseABI    = "failed to parse ABI file"
 	ErrOpenBINFile = "failed to open BIN file"
+	ErrNoABIInFile = "no ABI content not found in file"
```
skudasov, line 23: Nit, found in file?

seth/contract_store_test.go
```diff
@@ -10,40 +10,102 @@ import (
...
 		},
 		{
 			name:    "can't open the ABI path",
 			abiPath: "dasdsadd",
 			err:     "open dasdsadd: no such file or directory",
 		},
+		{
+			name:              "can't open the gethwrappers path",
+			gethWrappersPaths: []string{"dasdsadd"},
+			err:               "lstat dasdsadd: no such file or directory",
```
skudasov, line 51: Maybe wrap them and add more context, like `failed to find bind wrappers` or something? Perhaps it is already evident from logs but it won't hurt IMO

seth/contract_store_test.go
```diff
@@ -10,40 +10,102 @@ import (
...
+		},
 		{
 			name:    "empty ABI dir",
 			abiPath: "./contracts/emptyContractDir",
 		},
 		{
-			name:    "invalid ABI inside dir",
+			name:              "empty gethwrappers dir",
+			gethWrappersPaths: []string{"./contracts/emptyContractDir"},
```
skudasov, line 69: Maybe we should throw an error if the dir is empty? Misconfiguration can happen but I don't want to run without bindings, I usually have them

shell.nix
```diff
@@ -70,5 +69,6 @@ mkShell' {
     export PATH=$GOBIN:$PATH
     # install gotestloghelper
     go install github.com/smartcontractkit/chainlink-testing-framework/tools/gotestloghelper@latest
+    go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.57.2
```
skudasov, line 72: To make it clear to other people: it's a temporary solution. The proper solution is to upgrade `nix` lock files but we can't do that right now because of Geth versions incompatibility

seth/keyfile.go
```diff
@@ -13,22 +13,27 @@ import (
 )
 
 // NewAddress creates a new address
-func NewAddress() (string, string, error) {
-	privateKey, err := crypto.GenerateKey()
+func NewAddress() (address, privateKey string, err error) {
```
skudasov, line 16: Let's stick with one way of writing return params, let's use anonymous since most of our code use them this way, easier to read

seth/client_builder.go
```diff
@@ -227,6 +227,13 @@ func (c *ClientBuilder) WithArtifactsFolder(folder string) *ClientBuilder {
 	return c
 }
 
+// WithGethWrappersFolders sets list of folders where the Geth wrappers are stored. Seth will load ABIs from all wrappers it finds in theses folders (including subfolders).
+// Default value is an empty string (= loading disabled).
+func (c *ClientBuilder) WithGethWrappersFolders(folders []string) *ClientBuilder {
```
skudasov, line 232: Let's add it to our docs? https://github.com/smartcontractkit/chainlink-testing-framework/blob/tt_1511_load_abi_gethwrappers/seth/README.md

skudasov (changes requested):
LGTM 🚀 Some nits on errors and docs

skudasov (approved):

PR #1119, opened 2024-09-05
Make geth, eth2-common and besu charts compatible with staging cluster

scheibinger:
* Adjust geth-prysm, besu-prysm and eth-common charts to work with rootless pods in kubernetes.
  * Define security context 
  * use 1000 user instead of root for running container commands
  * Switch to updated docker image with genesis generator
* Upgrade dependencies to the latest versions

Required for CRIB-345

## Testing
* Tested both geth + prysm and besu + prysm on CRIB.
* Starting fresh deployment worked fine, restarting geth pod worked fine. Restarting beacon chain node didn't work as per existing limitations.


lib/charts/besu-prysm/templates/besu.statefulset.yaml
```diff
@@ -60,7 +52,6 @@ spec:
             - "--data-storage-format=BONSAI"
             - "--rpc-tx-feecap=0"
             - "--logging={{ .Values.excutionClient.logging.level }}"
-            - "--kzg-trusted-setup=/data/custom_config_data/trusted_setup.txt"
```
scheibinger, line 63: having this flag was causing an error: 

```
2024-09-13 15:05:32.947+00:00 | main | ERROR | Besu | Failed to start Besu
java.lang.RuntimeException: Couldn't load Trusted Setup. File might not exist or there is a permission issue.
	at ethereum.ckzg4844.CKZG4844JNI.loadTrustedSetup(Native Method)
	at org.hyperledger.besu.evm.precompile.KZGPointEvalPrecompiledContract.init(KZGPointEvalPrecompiledContract.java:63)
	at org.hyperledger.besu.cli.BesuCommand.configureNativeLibs(BesuCommand.java:1542)
	at org.hyperledger.besu.cli.BesuCommand.run(BesuCommand.java:1193)
	at picocli.CommandLine.executeUserObject(CommandLine.java:2026)
	at picocli.CommandLine.access$1500(CommandLine.java:148)
```
Not sure if that was related to my changes or the upgrade of besu to the newer version.
After removing the flag it started building blocks fine, and everything worked.

@Tofel Do we need that?

Tofel: path was incorrect, it should be `"--kzg-trusted-setup=/data/metadata/trusted_setup.txt"`

this param is required only by some Besu versions.

lib/charts/besu-prysm/Chart.lock
```diff
@@ -1,6 +1,6 @@
 dependencies:
 - name: eth2-common
-  repository: https://raw.githubusercontent.com/smartcontractkit/qa-charts/gh-pages/
-  version: 0.5.2
-digest: sha256:6bc7b0b10fab00cba405205673182004e440508efc06d51a80b6ddbf3f7300e1
-generated: "2024-03-07T13:05:57.276874+01:00"
+  repository: file://../eth2-common
+  version: 0.5.3
```
scheibinger, line 4: we need to run `helm dependency update` after eth2-common is already pushed to registry

Tofel (approved):

PR #1124, opened 2024-09-06
[TT- 1545] lower the amount of WARN and ERROR logs

Tofel:
This PR lowers ERR and WARN level logs to a manageable one and basically only uses that level if user can should take some action or it's a catastrophically failure. Everything else was switched to DEBUG.

---

## Why
The changes improve logging and error handling in various parts of the code, making it more appropriate for different scenarios. For instance, changing logging levels in certain situations from warning to debug can help in reducing unnecessary alarm or clutter in the log files under normal circumstances. Also, enhancing error messages with more details improves the ability to diagnose issues.

## What
- **seth/client.go**
  - Changed logging level from `Warn` to `Debug` for messages related to ephemeral mode and multiple RPC URLs to reduce unnecessary warnings.
  - Improved error logging in transaction waiting and nonce management to include more detailed information.
- **seth/config.go**
  - Changed logging level from `Warn` to `Debug` when no network name is provided and the default network is used, to make the log less alarming.
- **seth/contract_store.go**
  - Enhanced logging in ABI loading from Geth wrappers by including the error reason in the debug message when an ABI is not found in a file.
- **seth/decode.go**
  - Modified logging levels and messages for decoding errors and tracing errors to provide more context or to reduce unnecessary alarm.
- **seth/dot_graph.go**
  - Updated a warning message to include contact information for the Test Tooling team in case of unusual occurrences, improving support.
- **seth/gas_adjuster.go**
  - Improved error handling and logging for gas price adjustments and estimation, including more contextual information in debug logs.
- **seth/http_logging_transport.go**
  - Changed error logging to warning when failing to dump HTTP requests or responses, to better reflect the severity of the issue.
- **seth/retry.go**
  - Updated logging level to `Info` when bumping gas for a transaction, reflecting the action's importance without indicating an error.
- **seth/tracing.go**
  - Improved error handling in trace decoding by changing a log level to debug and adding more detailed error messages.

These changes collectively aim to improve the clarity of logs and error messages, making them more useful for debugging and less likely to cause unnecessary concern during normal operations.

### DIFF
--- a/seth/client.go
+++ b/seth/client.go
@@ -92,7 +92,7 @@ func NewClientWithConfig(cfg *Config) (*Client, error) {
 		// we don't care about any other keys, only the root key
 		// you should not use ephemeral mode with more than 1 key
 		if len(cfg.Network.PrivateKeys) > 1 {
-			L.Warn().Msg("Ephemeral mode is enabled, but more than 1 key is loaded. Only the first key will be used")
+			L.Debug().Msg("Ephemeral mode is enabled, but more than 1 key is loaded. Only the first key will be used")
 		}
 		cfg.Network.PrivateKeys = cfg.Network.PrivateKeys[:1]
 		pkeys, err := NewEphemeralKeys(*cfg.EphemeralAddrs)
@@ -230,7 +230,7 @@ func NewClientRaw(
 		return nil, errors.New("no RPC URL provided")
 	}
 	if len(cfg.Network.URLs) > 1 {
-		L.Warn().Msg("Multiple RPC URLs provided, only the first one will be used")
+		L.Debug().Msg("Multiple RPC URLs provided, only the first one will be used")
 	}
 	ctx, cancel := context.WithTimeout(context.Background(), cfg.Network.DialTimeout.Duration())
 	defer cancel()
@@ -307,7 +307,7 @@ func NewClientRaw(
 
 	if cfg.CheckRpcHealthOnStart {
 		if c.NonceManager == nil {
-			L.Warn().Msg("Nonce manager is not set, RPC health check will be skipped. Client will most probably fail on first transaction")
+			L.Debug().Msg("Nonce manager is not set, RPC health check will be skipped. Client will most probably fail on first transaction")
 		} else {
 			if err := c.checkRPCHealth(); err != nil {
 				return nil, err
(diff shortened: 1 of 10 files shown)


seth/client.go
```diff
@@ -92,7 +92,7 @@ func NewClientWithConfig(cfg *Config) (*Client, error) {
 		// we don't care about any other keys, only the root key
 		// you should not use ephemeral mode with more than 1 key
 		if len(cfg.Network.PrivateKeys) > 1 {
-			L.Warn().Msg("Ephemeral mode is enabled, but more than 1 key is loaded. Only the first key will be used")
+			L.Debug().Msg("Ephemeral mode is enabled, but more than 1 key is loaded. Only the first key will be used")
```
skudasov, line 95: Deserves to be at least `info`?

seth/client.go
```diff
@@ -230,7 +230,7 @@ func NewClientRaw(
 		return nil, errors.New("no RPC URL provided")
 	}
 	if len(cfg.Network.URLs) > 1 {
-		L.Warn().Msg("Multiple RPC URLs provided, only the first one will be used")
+		L.Debug().Msg("Multiple RPC URLs provided, only the first one will be used")
```
skudasov, line 233: Deserves to be at least `info`?

seth/gas_adjuster.go
```diff
@@ -235,18 +234,15 @@ func (m *Client) GetSuggestedEIP1559Fees(ctx context.Context, priority string) (
...
 			Float64("BaseFee", baseFee64).
 			Int64("SuggestedTip", currentGasTip.Int64()).
-			Msg("Incorrect gas data received from node. Skipping automation gas estimation")
+			Msgf("Incorrect gas data received from node: historical base fee was 0. Skipping automation gas estimation")
 		return
 	}
 
 	if currentGasTip.Int64() == 0 {
-		L.Warn().
+		L.Debug().
```
skudasov, line 245: If it can cause a transaction to hang indefinitely from your experience maybe we better still have it as `Warn`?

seth/gas_adjuster.go
```diff
@@ -491,15 +483,15 @@ func (m *Client) HistoricalFeeData(priority string) (baseFee float64, historical
 		historicalGasTipCap = stats.TipCap.Perc25
 	default:
 		err = fmt.Errorf("unknown priority: %s", priority)
-		L.Error().
+		L.Debug().
 			Str("Priority", priority).
-			Msg("Unknown priority. Skipping automation gas estimation")
+			Msgf("Unknown priority: %s. Skipping automation gas estimation", err.Error())
```
skudasov, line 488: Can we enforce it by type so you can't insert a "string"?

skudasov (approved):

seth/gas_adjuster.go
Tofel: yup, in that case let me use fallback value from config

seth/gas_adjuster.go
Tofel: we can. but would need to experiment how that would work with TOML

seth/client.go
Tofel: set to warn

seth/client.go
Tofel: set to warn

PR #1126, opened 2024-09-06
[TT-1619] more robust builder and better logging

Tofel:
Changes:
* more robust client builder that will panic with our message instead of NPE in the unlikely case that `Network` is `nil`
* when modifying `Network` functions will now iterate over `Networks` to find a matching one by name instead of modifying the first one
* more human-friendly logging of gas bumping values

---

## Why
The changes improve transaction handling and configuration management within the Ethereum client implementation. Enhancements include better logging for transaction mining, streamlined configuration setting for network-specific parameters, and more robust gas price bumping strategies for transactions awaiting confirmation. These updates aim to increase the client's reliability, ease of configuration, and adaptability to network conditions.

## What
- **seth/client.go**
  - Added logging to indicate when the system is waiting for a transaction to be mined, enhancing visibility into the transaction lifecycle.
  - Removed redundant transaction hash logging in several places to clean up log output.
- **seth/client_builder.go**
  - Introduced `assertNetworkIsSet` method calls in builder methods to ensure network configuration is initialized before setting specific network options, improving the robustness of client configuration.
  - Replaced direct slice modifications with `findNetworkByName` method calls for setting network-specific configurations, making the code more maintainable and readable.
  - Added `assertNetworkIsSet` call and parameter renaming in `WithTransferGasFee` for consistency and to enforce network configuration initialization.
- **seth/config.go**
  - Added `findNetworkByName` method to retrieve network configurations by name, facilitating more dynamic and readable network configuration management.
- **seth/retry.go**
  - Enhanced gas price bumping logging with more detailed information, including old/new prices and their difference in both wei and ether, improving debugging and monitoring of gas bumping processes.
  - Replaced direct gas price modifications with calculated new prices and checks against max gas price, making the gas bumping strategy more adaptable and safer against exceeding predefined gas price limits.
  - Adjusted transaction type handling in `prepareReplacementTransaction` to ensure proper gas price bumping and transaction creation based on transaction type, enhancing support for multiple transaction types and their specific needs.

### DIFF
--- a/seth/client.go
+++ b/seth/client.go
@@ -483,6 +483,9 @@ func (m *Client) TransferETHFromKey(ctx context.Context, fromKeyNum int, to stri
 
 // WaitMined the same as bind.WaitMined, awaits transaction receipt until timeout
 func (m *Client) WaitMined(ctx context.Context, l zerolog.Logger, b bind.DeployBackend, tx *types.Transaction) (*types.Receipt, error) {
+	l.Info().
+		Str("TX", tx.Hash().String()).
+		Msg("Waiting for transaction to be mined")
 	queryTicker := time.NewTicker(time.Second)
 	defer queryTicker.Stop()
 	ctx, cancel := context.WithTimeout(ctx, m.Cfg.Network.TxnTimeout.Duration())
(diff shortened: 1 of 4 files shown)


seth/client.go
```diff
@@ -483,6 +483,9 @@ func (m *Client) TransferETHFromKey(ctx context.Context, fromKeyNum int, to stri
 
 // WaitMined the same as bind.WaitMined, awaits transaction receipt until timeout
 func (m *Client) WaitMined(ctx context.Context, l zerolog.Logger, b bind.DeployBackend, tx *types.Transaction) (*types.Receipt, error) {
+	l.Info().
```
skudasov, line 486: Can it be spammy, we are using it inside our functions, right?

skudasov (approved):

seth/client.go
Tofel: yes, but it will run more than once only if gas bumping is enabled.

PR #1141, opened 2024-09-18
Add base_url_gap for Grafana

lukaszcl:
Related PR in core: https://github.com/smartcontractkit/chainlink/pull/14475

---

## Why
The changes enhance Grafana logging configuration and log stream handling, particularly for CI environments. By introducing a new configuration parameter for Grafana to specify a base URL for dashboard access through a GAP proxy, it simplifies accessing dashboards from CI pipelines. Furthermore, the adjustment in log stream handlers to utilize this new URL for generating dashboard links ensures that the links are correctly generated for environments where direct access to Grafana might be restricted.

## What
- **lib/config/logging.go**
  - Added `BaseUrlGap` field to `GrafanaConfig` struct. This addition allows specifying a base URL for the Grafana dashboard through a GAP proxy, primarily used in CI environments.
- **lib/logstream/logstream_handlers.go**
  - Modified `GetLogLocation` method to use `BaseUrlGap` from `GrafanaConfig` if available. This change ensures that the generated Grafana dashboard URL is accessible in environments where a GAP proxy is needed for dashboard access.


lib/config/logging.go
```diff
@@ -91,6 +91,7 @@ func (l *LokiConfig) Validate() error {
 
 type GrafanaConfig struct {
 	BaseUrl      *string `toml:"base_url"`
+	BaseUrlCI    *string `toml:"base_url_github_ci"` // Base URL for the dashboard via GAP proxy used on CI
```
Tofel, line 94: `// Base URL for the dashboard via GAP proxy used on CI` -> `// URL of GAP proxy used on CI for Grafana`

lib/logstream/logstream_handlers.go
```diff
@@ -207,17 +199,31 @@ func (h *LokiLogHandler) GetLogLocation(consumers map[string]*ContainerLogConsum
...
+
+		// try to shorten the URL only if we have all the required configuration parameters
+		shortUrl, err := ShortenUrl(baseUrl, sb.String(), *h.loggingConfig.Grafana.BearerToken)
+		if err != nil {
+			return "", err
+		}
+		h.grafanaUrl = shortUrl
+	} else {
+		if h.loggingConfig.Grafana == nil || h.loggingConfig.Grafana.BaseUrl == nil {
+			return "", errors.New("grafana base URL is not set in logging config")
```
Tofel, line 219: I don't think we need to return error. Why not display what we can? which is the part of url without hostname? in that way if you forget to set it in the config you will still have an usable url like: `/dashboard-ulr&var=1...` so you can just manually paste it after Grafana URL and get a working link

Tofel (approved):

PR #1203, opened 2024-10-03
[ship-3765]Adds wrapper for Loki Rest api to query logs

davidcauchi:
---

## Why
The changes introduce a new Loki client for querying logs from Loki, a horizontally-scalable, highly-available, multi-tenant log aggregation system inspired by Prometheus. The client facilitates interactions with Loki to fetch and process log data, making it easier to integrate log querying capabilities within applications.

## What
- **lib/client/loki.go**
  - Added `LokiResponse` struct to model the structure of responses from Loki.
  - Added `LokiLogEntry` struct to represent a single log entry.
  - Added `LokiClient` struct with fields for the base URL, tenant ID, basic auth credentials, and query parameters.
  - Added `LokiQueryParams` struct to hold parameters required for querying Loki.
  - Implemented `NewLokiClient` function to create a new instance of `LokiClient`.
  - Implemented `QueryLogs` method for `LokiClient` to query logs based on provided parameters and return raw log entries.
  - Added private method `extractRawLogEntries` to process the Loki response and extract log entries.

### DIFF
--- a/lib/client/loki.go
+++ b/lib/client/loki.go
@@ -11,6 +11,23 @@ import (
 	"github.com/smartcontractkit/chainlink-testing-framework/lib/logging"
 )
 
+// LokiAPIError is a custom error type for handling non-200 responses from the Loki API
+type LokiAPIError struct {
+	StatusCode int
+	Message    string
+}
+
+// Implement the `Error` interface for LokiAPIError
+func (e *LokiAPIError) Error() string {
+	return fmt.Sprintf("Loki API error: %s (status code: %d)", e.Message, e.StatusCode)
+}
+
+// LokiBasicAuth holds the authentication details for Loki
+type LokiBasicAuth struct {
+	Username string
+	Password string
+}
+
 // LokiResponse represents the structure of the response from Loki
 type LokiResponse struct {
 	Data struct {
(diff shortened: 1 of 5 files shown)


lib/client/loki.go
```diff
@@ -0,0 +1,155 @@
...
+		"start": lc.QueryParams.StartTime.Format(time.RFC3339Nano),
+		"end":   lc.QueryParams.EndTime.Format(time.RFC3339Nano),
+		"limit": fmt.Sprintf("%d", lc.QueryParams.Limit),
+	}
+
+	// Send request using the pre-configured Resty client
+	resp, err := lc.RestyClient.R().
+		SetContext(ctx).
+		SetHeader("X-Scope-OrgID", lc.TenantID).
+		SetBasicAuth(lc.BasicAuth, "").
```
skudasov, line 99: Not sure we can send `login:password` as one string here. Either let's remove it for now or split the field and insert `Login` and `Password` properly.

skudasov (changes requested):

lib/client/loki.go
davidcauchi: This has been split.

skudasov (approved):

PR #1360, opened 2024-11-19
[TT-1722] WASP new docs as mdbook

Tofel:
Also:
* fixed API token generation to work with latest Grafana and service accounts
* small streamlining of `Profile.Run()`
* some new examples


book/src/libs/wasp/components/generator.md
```diff
@@ -0,0 +1,57 @@
...
+> RPS load type can only be used with a `Gun`, while VUs can only be used with a `VirtualUser`.
+
+---
+
+### Choosing Between `Gun` and `VirtualUser`
+
+#### **`Gun`**
+* Best for **stateless protocols** (e.g., HTTP).
+* Simplistic in nature; ideal for executing a single operation that does not require setup or teardown.
+* Operates using a **closed model**, where:
```
skudasov, line 23: It is an `open` model, if the service can't process our `N RPS` we'll use more goroutines to test that it can scale strictly to our required workload of `N RPS`
Like `Google`, it is `open` no one can prohibit you to use `Google`

book/src/libs/wasp/components/generator.md
```diff
@@ -0,0 +1,57 @@
...
+  * The number of requests is fixed.
+  * The load adjusts to meet the target RPS, regardless of the system's response time.
+* Recommended for scenarios focused on measuring **throughput**.
+
+#### **`VirtualUser`**
+* Designed for **stateful protocols** (e.g., `WebSocket`) or workflows involving multiple operations (e.g., authenticating, executing tasks, and logging out).
+* More complex, with dedicated methods for setup and teardown.
+* Operates using an **open model**, where:
```
skudasov, line 31: This is a `closed` model. We model the user and if all `VUs` will be stuck in waiting we won't generate more load. The model is `closed`, like `SWIFT`, we do not expect internal DDoS from internal users
Please also change the explanation below, `open` and `closed`

skudasov: To explain it very simply, when we test an `open` system, we do not expect a load feedback loop. We generate the load no matter what because the system is `open` for any user.
In `closed` systems amount of users are usually limited and controlled so we use `VU` and if the system can't handle more all `VUs` will sleep, we won't generate new virtual users
@Tofel

skudasov (changes requested):

book/src/libs/wasp/profile_test.md
```diff
@@ -0,0 +1,108 @@
...
+You can find the full example [here](https://github.com/smartcontractkit/chainlink-testing-framework/tree/main/wasp/examples/profiles).
+
+---
+
+> [!NOTE]  
+> The `error` returned by the `.Run(true)` method only indicates if any dashboard alerts were triggered.  
+> To check for load generation errors, call the `Errors()` method on each `Generator` in the `Profile`.
+
+> [!NOTE]  
+> Currently, it’s not possible to have a "waiting" schedule that doesn’t generate any load. To implement such logic, start one `Profile` in a goroutine and use `time.Sleep` before starting the second `Profile`. An example of this can be found [here](https://github.com/smartcontractkit/chainlink-testing-framework/tree/main/wasp/examples/profiles/node_background_load_test.go).
```
skudasov, line 102: You can use `Run(false)` to not wait for the profile, you don't need to run it inside the goroutine. I'd say we remove it for now, add a case for 0 RPS and then bring the explanation back.

book/src/libs/wasp/components/profile.md
```diff
@@ -0,0 +1,15 @@
+# WASP - Profile
+
+A **Profile** allows you to combine load from different generators. Each generator operates according to its own schedule, and you can even mix RPS and VU load types, as shown in [this example](https://github.com/smartcontractkit/chainlink-testing-framework/blob/main/wasp/examples/profiles/node_mixed_test.go).
+
+> [!NOTE]  
+> The `error` returned by the `.Run(true)` function of a `Profile` only indicates if any dashboard alerts were triggered.
```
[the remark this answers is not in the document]
Tofel: I should probably make that comment more specific, because if you did use `.WithGrafana(opts *GrafanaOpts)` and your dashboard had alerts sets and any of them was triggered, we do return an error:
```
	if m.grafanaOpts.CheckDashboardAlertsAfterRun != "" {
		m.printDashboardLink()
		alerts, err := CheckDashboardAlerts(m.grafanaAPI, m.startTime, time.Now(), m.grafanaOpts.CheckDashboardAlertsAfterRun)
		if len(alerts) > 0 {
			log.Info().Msgf("Alerts found\n%s", grafana.FormatAlertsTable(alerts))
		}
		if err != nil {
			return m, err
		}
```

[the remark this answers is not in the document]
Tofel: added extra explanation

book/src/libs/wasp/components/generator.md
Tofel: yeah, mixed the names, not how they work 🫨

skudasov (approved):

book/src/libs/wasp/k8s.md
```diff
@@ -0,0 +1,3 @@
+# WASP - k8s
+
+TBD
```
AKhozya, line 3: Just checking that this is not lost in action

book/src/libs/wasp/profile_test.md
```diff
@@ -0,0 +1,112 @@
...
+
+### Background Load Schedule for the `Gun`
+
+For the RPS `Gun`, we’ll define a schedule where:
+1. During the first 10 seconds, the RPS increases by 2.5 every second.
+2. For the next 40 seconds, it remains steady at 5 RPS.
+
+```go
+epsilonSchedule := wasp.Combine(
+    wasp.Steps(1, 1, 4, 10*time.Second), // Start at 1 RPS, increment by 1 RPS in 4 steps over 10 seconds (1 increment every 2.5 seconds)
```
AKhozya, line 40: Named params will help a lot here and below

wasp/Makefile
```diff
@@ -34,7 +34,8 @@ dashboard:
 .PHONY: start
 start:
 	docker compose -f compose/docker-compose.yaml up -d
-	sleep 5 && curl -X POST -H "Content-Type: application/json" -d '{"name":"test", "role": "Admin"}' http://localhost:3000/api/auth/keys | jq .key
+	sleep 5
```
AKhozya, line 37: Wonder if this should be replaced with proper waiting, e.g. for log message or process up

AKhozya (changes requested):

book/src/libs/wasp/profile_test.md
AKhozya: ah, cool - I thought they come for free - my kotlin and ruby past :P

AKhozya (approved):

PR #1365, opened 2024-11-20
TT-1856 Fix err consumer has finished but you are still trying to accept logs

lukaszcl:
---

## Why
The changes focus on error handling, logging behavior adjustments, and managing log stream consumer lifecycle more efficiently. Specifically, the changes aim to downgrade certain errors to warnings to avoid overly aggressive error logs during test execution, introduce checks to gracefully handle log production errors when containers are not running, implement a safer way to stop log consumers using `sync.Once` to ensure idempotency, and improve the logic for handling log consumer disconnection and stopping.

## What
- **lib/logging/log.go**
  - Changed logging level from `Error` to `Warn` for messages after test ended to reduce the severity of these log messages.
- **lib/logstream/logstream.go**
  - Added a check to skip log producer error handling if the container is not running, preventing unnecessary error logs.
  - Introduced `sync.Once` in the `stop()` method of `ContainerLogConsumer` to ensure the stop logic is executed only once, preventing potential race conditions or multiple executions.
  - Modified `stop()` to signal the log listening channel and close it safely, ensuring a cleaner shutdown process.
  - Added `IsStopped()` to check if the log consumer has been stopped, enhancing the control flow and making the consumer lifecycle management more explicit.
  - Changed `DisconnectContainer` to always log a disconnection attempt, ensuring consistency in log output.
  - Adjusted the `FlushLogsToTargets` function to use the new `stop()` method, aligning with the updated consumer stopping mechanism.
  - Removed `isDone` flag usage in favor of the new `stop()` method and `IsStopped()` check, streamlining the consumer's state management.

### DIFF
--- a/lib/logstream/logstream.go
+++ b/lib/logstream/logstream.go
@@ -233,7 +233,7 @@ func (m *LogStream) ConnectContainer(ctx context.Context, container LogProducing
 					if !container.IsRunning() {
 						m.log.Info().
 							Str("Container name", name).
-							Msg("Skipping log producer error as the container is not running anymore")
+							Msg("Skipping log producer retrying, because the container is not running anymore")
 						break
 					}
 


lib/logstream/logstream.go
```diff
@@ -229,6 +229,14 @@ func (m *LogStream) ConnectContainer(ctx context.Context, container LogProducing
 			select {
 			case logErr := <-container.GetLogProductionErrorChannel():
 				if logErr != nil {
+					// Check if the container is not stopped or terminated
+					if !container.IsRunning() {
+						m.log.Info().
+							Str("Container name", name).
+							Msg("Skipping log producer error as the container is not running anymore")
```
Tofel, line 236: `Skipping log producer retrying, because the container is not running anymore`.

lib/logstream/logstream.go
```diff
@@ -784,12 +809,8 @@ func (g *ContainerLogConsumer) Accept(l tc.Log) {
 		return
 	}
 
-	if g.isDone {
-		g.ls.log.Error().
-			Str("Test", g.ls.testName).
-			Str("Container", g.name).
-			Str("Log", string(l.Content)).
-			Msg("Consumer has finished, but you are still trying to accept logs. This should never happen")
```
Tofel, line 792: how will you know now that you're tying to send the logs, even though you should not?

lukaszcl: @Tofel to be honest I don't care about it here. I think this can still happen until we migrate to "Passing the LogConsumers in the ContainerRequest" approach https://golang.testcontainers.org/features/follow_logs/?utm_source=chatgpt.com#passing-the-logconsumers-in-the-containerrequest. This will be a separate ticket.

lib/logstream/logstream.go
lukaszcl: @Tofel fixed

Tofel (approved):

PR #1506, opened 2024-12-19
use partial gas price percentiles

Tofel:
if some percentile calculations fail due to insufficient data we will still try to use what we could calculate instead of erroring

---

## Why
The changes improve the robustness and functionality of gas estimation in the blockchain client application by handling partial or incorrect gas estimation data more gracefully and enhancing error handling and logging.

## What
- **seth/client.go**
  - Improved error handling by considering "Partial data received" as a condition to disable gas price estimations and fallback to hardcoded values.
  - Enhanced logging messages when receiving incorrect gas estimations or when EIP1559 fees are not supported by the network, including conditions where the gas price is set to 0.
- **seth/gas.go**
  - Modified error handling in the `Stats` function to log errors and continue instead of returning immediately upon encountering errors calculating gas percentiles.
  - In `quantilesFromFloatArray`, replaced direct error returns with logging the error and continuing, facilitating the use of partial data for gas estimation.
- **seth/gas_adjuster.go**
  - Introduced checks for NaN values in gas price and tip cap calculations to handle partial data scenarios more effectively. It logs an error and skips automation gas estimation in case of NaN values, indicating partial data reception.

### DIFF
--- a/seth/client.go
+++ b/seth/client.go
@@ -934,7 +934,7 @@ func (m *Client) CalculateGasEstimations(request GasEstimationRequest) GasEstima
 	defer cancel()
 
 	var disableEstimationsIfNeeded = func(err error) {
-		if strings.Contains(err.Error(), ZeroGasSuggestedErr) {
+		if strings.Contains(err.Error(), ZeroGasSuggestedErr) || strings.Contains(err.Error(), "Partial data received") {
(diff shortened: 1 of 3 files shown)


Tofel:
Superseded by https://github.com/smartcontractkit/chainlink-testing-framework/pull/1532

PR #1519, opened 2024-12-30
[DPA-1458]: feat(solana): auto deploy programs

graham-chainlink:
By using `--upgradeable-program` , we can auto deploy programs detected in the mounted path `/programs`, this reduces the need for consumer of this library to have to deploy the programs themselves, especially with go sdk where deploying a program is not straightforward and involves a decent amount of code, the alternative is to ssh into the docker container and run `solana deploy ...` but this involves an extra step.

Inspired from https://github.com/smartcontractkit/chainlink-ccip/blob/609f7b0c9734b3cfc1d1a1aea0fd1ddf4c90bd0c/chains/solana/contracts/tests/testutils/anchor.go#L72-L72

JIRA: https://smartcontract-it.atlassian.net/browse/DPA-1458

---

## Why
The changes introduce a new feature to deploy upgradeable Solana programs at the start of the Solana test validator. This enhancement allows users to specify a map of Solana program names to their corresponding program IDs directly in the blockchain configuration, facilitating a more dynamic and flexible test setup.

## What
- **framework/components/blockchain/blockchain.go**
  - Added `SolanaPrograms map[string]string` to the `Input` struct to allow specifying Solana programs to deploy upon test validator start.
- **framework/components/blockchain/solana.go**
  - Introduced a new logic to construct the `solana-test-validator` command with flags for deploying specified Solana programs. This includes iterating over the `SolanaPrograms` map from the configuration and appending program deployment flags to the validator startup command.
  - Modified the container entrypoint to utilize the newly constructed command with program deployment flags, facilitating the deployment of specified programs when the Solana test validator starts.

### DIFF
--- a/book/src/framework/components/blockchains/solana.md
+++ b/book/src/framework/components/blockchains/solana.md
@@ -17,10 +17,14 @@ arm64 f4hrenh9it/solana:latest - used locally
   # optional, in case you need some custom image
   # image = "solanalabs/solana:v1.18.26"
 
-# optional, in case you need to deploy some programs
-# this example assumes there is a mcm.so in the mounted contracts_dir directory
-# value is the program id to set for the deployed program instead of auto generating
-# useful for deterministic testing
+# optional
+# To deploy a solana program, we can provide a mapping of program name to program id.
+# The program name has to match the name of the .so file in the contracts_dir directory.
+# This example assumes there is a mcm.so in the mounted contracts_dir directory
+# value is the program_id to set for the deployed program, any valid public key can be used here.
+# this allows lookup via program_id during testing.
+# Alternative, we can use `solana deploy` instead of this field to deploy a program, but 
+# program_id will be auto generated.
 [blockchain_a.solana_programs]
   mcm = "6UmMZr5MEqiKWD5jqTJd1WCR5kT8oZuFYBLJFi1o6GQX"
 
--- a/book/src/framework/components/blockchains/solana.md
+++ b/book/src/framework/components/blockchains/solana.md
@@ -16,6 +16,14 @@ arm64 f4hrenh9it/solana:latest - used locally
   contracts_dir = "."
   # optional, in case you need some custom image
   # image = "solanalabs/solana:v1.18.26"
+
+# optional, in case you need to deploy some programs
+# this example assumes there is a mcm.so in the mounted contracts_dir directory
+# value is the program id to set for the deployed program instead of auto generating
+# useful for deterministic testing
+[blockchain_a.solana_programs]
+  mcm = "6UmMZr5MEqiKWD5jqTJd1WCR5kT8oZuFYBLJFi1o6GQX"
+
 ```
 
 ## Usage
--- a/framework/components/blockchain/blockchain.go
+++ b/framework/components/blockchain/blockchain.go
@@ -23,6 +23,7 @@ type Input struct {
 	ContractsDir string `toml:"contracts_dir"`
 	// programs to deploy on solana-test-validator start
 	// a map of program name to program id
+	// there needs to be a matching .so file in contracts_dir
 	SolanaPrograms map[string]string `toml:"solana_programs"`
 }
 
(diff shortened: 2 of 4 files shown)


framework/components/blockchain/blockchain.go
```diff
@@ -21,6 +21,10 @@ type Input struct {
 	// publickey to mint when solana-test-validator starts
 	PublicKey    string `toml:"public_key"`
 	ContractsDir string `toml:"contracts_dir"`
+	// programs to deploy on solana-test-validator start
+	// a map of program name to program id
+	// there needs to be a matching .so file in contracts_dir
```
Tofel, line 26: matching meaning `<program_name.so>`?

graham-chainlink: Yeah, 
```
[solana_config.solana_programs]
mcm = "6UmMZr5MEqiKWD5jqTJd1WCR5kT8oZuFYBLJFi1o6GQX"
```

expects the file `<contracts_dir>/mcm.so` and it will use `6UmMZr5MEqiKWD5jqTJd1WCR5kT8oZuFYBLJFi1o6GQX` as the program id , so it can be looked up later for testing.

> also, should programs reside in the root of contracts_dir or in <contracts_dir>/programs?

No strong opinions, i think contracts are programs in solana, but i was basically following the convention in [chainlink-ccip](https://github.com/smartcontractkit/chainlink-ccip/blob/609f7b0c9734b3cfc1d1a1aea0fd1ddf4c90bd0c/chains/solana/contracts/tests/testutils/anchor.go#L72-L72)

book/src/framework/components/blockchains/solana.md
```diff
@@ -16,6 +16,14 @@ arm64 f4hrenh9it/solana:latest - used locally
   contracts_dir = "."
   # optional, in case you need some custom image
   # image = "solanalabs/solana:v1.18.26"
+
+# optional, in case you need to deploy some programs
+# this example assumes there is a mcm.so in the mounted contracts_dir directory
+# value is the program id to set for the deployed program instead of auto generating
+# useful for deterministic testing
+[blockchain_a.solana_programs]
+  mcm = "6UmMZr5MEqiKWD5jqTJd1WCR5kT8oZuFYBLJFi1o6GQX"
```
Tofel, line 25: do you think you might expand that topic a bit? for example:
* how should one load a program with auto-generated id?
* where should one get that program id from, if auto-generation is not desired?
* in what way is testing deterministic with static program id? what indeterminism there would be if I used different program ids for programs that are the same, i.e. with the same bytecode? or is program id derived from the bytecode somehow?

it's more than fine to just outline these topics in broad strokes and linking to resources which describe them in more details

Tofel (changes requested):
if possible, please improve documentation (see comment), otherwise 👍

book/src/framework/components/blockchains/solana.md
graham-chainlink: Thanks, 

> how should one load a program with auto-generated id?

then we do it the normal way, we just do [solana deploy program](https://solana.com/docs/programs/deploying#deploy-program) then the program will be deployed with its own program id.

> where should one get that program id from, if auto-generation is not desired?

basically just any public key, but usually the developer of the solana program will have one set , eg [here for mcm](https://github.com/smartcontractkit/chainlink-ccip/blob/609f7b0c9734b3cfc1d1a1aea0fd1ddf4c90bd0c/chains/solana/contracts/programs/mcm/src/lib.rs#L3-L3)

> in what way is testing deterministic with static program id? what indeterminism there would be if I used different program ids for programs that are the same, i.e. with the same bytecode? or is program id derived from the bytecode somehow?

basically during testing, we want to look up the deployed program based on the program id,
eg
```
data, accErr := s.SolanaClient.GetAccountInfoWithOpts(ctx, <PROGRAM_ID>, &rpc.GetAccountInfoOpts{
		Commitment: config.DefaultCommitment,
	})
```

If we dont tell solana to deploy a program with a certain program id, then we wont be able to look it up easily during testing since it is auto generated.

If we tell solana to deploy a program with a provided program id, then we can just look it up easily during testing.

Hope this make sense?

graham-chainlink: I added more info  to the `solana.md` , let me know if that is enough or you prefer it to be in other places too

Tofel (approved):

PR #1536, opened 2025-01-08
Fix PortMappings

archseer:
We were mapping `8080/tcp` to `8080/tcp`, which is technically invalid syntax but docker accepted it. We should map `8080/tcp` to `8080` (no protocol specifier). This fixes running CTF with podman.

The second change is exposing the testcontainer object from the solana component. We're programatically using it within a test environment without a TOML config and would like to be able to call `testcontainers.CleanupContainer(c)`

### DIFF
--- a/framework/components/blockchain/blockchain.go
+++ b/framework/components/blockchain/blockchain.go
@@ -2,6 +2,7 @@ package blockchain
 
 import (
 	"fmt"
+	"github.com/testcontainers/testcontainers-go"
 )
 
 // Input is a blockchain network configuration params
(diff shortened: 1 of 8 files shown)


framework/components/blockchain/blockchain.go
```diff
@@ -29,11 +30,12 @@ type Input struct {
...
 type Output struct {
-	UseCache      bool    `toml:"use_cache"`
-	Family        string  `toml:"family"`
-	ContainerName string  `toml:"container_name"`
-	ChainID       string  `toml:"chain_id"`
-	Nodes         []*Node `toml:"nodes"`
+	UseCache      bool                     `toml:"use_cache"`
+	Family        string                   `toml:"family"`
+	ContainerName string                   `toml:"container_name"`
+	Container     testcontainers.Container `toml:"-"`
```
graham-chainlink, line 36: I noticed this was only set for solana, do we want to do the other block chain too?

smickovskid (approved):

graham-chainlink (approved):

PR #1537, opened 2025-01-08
[ship-3971] Poll for headers when using http rpc

davidcauchi:
- Poll for headers when using http rpc

---

## Why
The changes introduce enhancements and new features to the blockchain client, focusing on improving header subscriptions and handling for Ethereum clients. These modifications aim to ensure that the client can adapt to different network capabilities, specifically between subscription-supported networks and those requiring polling for new headers. Additionally, an initialization process for header subscriptions is introduced, aiming to streamline how the client starts listening for new headers, whether through subscriptions or polling.

## What
- **`lib/.changeset/v1.50.21.md`**
  - Fixed formatting issue by ensuring a newline at the end of the file.
- **`lib/blockchain/blockchain.go`**
  - Added a new interface method `InitializeHeaderSubscription()` to the `EVMClient` interface to support initializing header subscriptions.
- **`lib/blockchain/ethereum.go`**
  - Introduced a new structure `SharedHeaderPoller` and a global registry `chainPollingRegistry` to manage shared polling across instances for networks that do not support subscriptions.
  - Modified `newEVMClient` to replace direct subscription initialization with a call to `InitializeHeaderSubscription()`, which decides between using subscriptions or polling based on the client's capabilities.
  - Implemented `InitializeHeaderSubscription()` in `EthereumClient` to initialize header subscriptions or start polling for headers if subscriptions are not supported. This includes setting up a shared polling mechanism for networks without subscription support.
  - Added `startHeaderPolling()` in `EthereumClient` to start a polling loop that fetches new headers at regular intervals if the network does not support subscriptions. This method also includes logic to prevent concurrent polling from multiple instances on the same network.
  - Updated `WaitForFinalizedTx()` to include a fallback polling mechanism for transaction finalization checking if subscriptions are not supported.
  - Implemented `InitializeHeaderSubscription()` in `EthereumMultinodeClient` to delegate the initialization of header subscriptions or polling to its underlying default client.


joaoluisam (approved):
LGTM

lib/blockchain/ethereum.go
```diff
@@ -1220,6 +1256,71 @@ func (e *EthereumClient) AvgBlockTime(ctx context.Context) (time.Duration, error
...
+	e.subscriptionWg.Add(1)
+
+	latestHeader, err := e.HeaderByNumber(context.Background(), nil) // Fetch the latest header
+	if err != nil {
+		return fmt.Errorf("failed to fetch the latest header during initialization: %w", err)
+	}
+	go func() {
+		defer e.subscriptionWg.Done()
+		// Initialize lastHeaderNumber dynamically
+		lastHeaderNumber := latestHeader.Number.Uint64() - 1
```
Tofel, line 1282: why `-1`? in the first run we could could just execute `e.receiveHeader(latestHeader)` after fetching last header.
won't doing `-1` always result in the latest block being fetched 2x in every subsequent run?

lib/blockchain/ethereum.go
```diff
@@ -1220,6 +1256,71 @@ func (e *EthereumClient) AvgBlockTime(ctx context.Context) (time.Duration, error
...
+
+					// Process headers from (lastHeaderNumber + 1) to latestHeader.Number
+					// We may need to add a rate limiter, if we run into issues.
+					for blockNum := lastHeaderNumber + 1; blockNum <= latestHeader.Number.Uint64(); blockNum++ {
+						header, err := e.HeaderByNumber(context.Background(), big.NewInt(int64(blockNum)))
+						if err != nil {
+							e.l.Error().Err(err).Uint64("BlockNumber", blockNum).Msg("Error fetching header during range processing")
+							continue
+						}
+						lastHeaderNumber = header.Number.Uint64()
```
Tofel, line 1306: why do we need it? maybe to be safe we could run `lastHeaderNumber = latestHeader.Number.Uint64()` after the for loop?

lib/blockchain/ethereum.go
```diff
@@ -1220,6 +1256,71 @@ func (e *EthereumClient) AvgBlockTime(ctx context.Context) (time.Duration, error
...
+				if latestHeader.Number.Uint64() > lastHeaderNumber {
+					lastHeaderNumber = latestHeader.Number.Uint64()
+					if err := e.receiveHeader(latestHeader); err != nil {
+						e.l.Error().Err(err).Msg("Error processing header")
+						continue
+					}
+
+					// Process headers from (lastHeaderNumber + 1) to latestHeader.Number
+					// We may need to add a rate limiter, if we run into issues.
+					for blockNum := lastHeaderNumber + 1; blockNum <= latestHeader.Number.Uint64(); blockNum++ {
```
Tofel, line 1300: will that work? in #1292 you have set `lastHeaderNumber` to `latestHeader` and here you are iterating from `latestHeader+1` to `latestHeader`, so this for loop shouldn't run even once...

maybe better just iterate like this:
```
blocksToFetch := latestHeader.Number.Uint64() - lastHeaderNumber
for blockNum := 0; blockNum < blocksToFetch; blockNum++ {
    header, err := e.HeaderByNumber(context.Background(), big.NewInt(int64(blockNum)))
    if err != nil {
        e.l.Error().Err(err).Uint64("BlockNumber", blockNum).Msg("Error fetching header during range processing")
	continue
    }
    if err := e.receiveHeader(header); err != nil {
        e.l.Error().Err(err).Uint64("BlockNumber", blockNum).Msg("Error processing header")
        continue
    }
}

lastHeaderNumber = latestHeader.Number.Uint64()
```
I would also use `<` as condition, because the last block would have already been fetched and added in lines #1293-1295?

davidcauchi: Good point. Please check the updated code.

lib/blockchain/ethereum.go
davidcauchi: Yes I had to remove one of these. Thanks

lib/blockchain/ethereum.go
```diff
@@ -1220,6 +1277,125 @@ func (e *EthereumClient) AvgBlockTime(ctx context.Context) (time.Duration, error
...
+	}
+	e.sharedPoller = poller
+	chainPollingRegistry.Unlock()
+	return e.startHeaderPolling()
+}
+
+// startPollingHeaders starts a polling loop to fetch new headers at regular intervals
+func (e *EthereumClient) startHeaderPolling() error {
+	e.l.Info().Int64("Chain Id", e.GetChainID().Int64()).Msg("Starting Polling Headers")
+	// pollInterval := time.Second * 30
```
Tofel, line 1306: remove?

lib/blockchain/ethereum.go
```diff
@@ -1220,6 +1277,125 @@ func (e *EthereumClient) AvgBlockTime(ctx context.Context) (time.Duration, error
...
+	chainPollingRegistry.Lock()
+	pollInterval := chainPollingRegistry.m[e.GetChainID().Int64()].pollInterval
+	ticker := time.NewTicker(pollInterval)
+	chainPollingRegistry.Unlock()
```
Tofel, line 1310: why do we need a lock here? to protect from concurrent read/writes? if so, then maybe you could use RWMutex? and here just use read lock, and where you wrote to `chainPollingRegistry` use a write lock?

Tofel (approved):

PR #1538, opened 2025-01-08
TT-1911 Fix 0% flakiness if ratio if only few tests were flaky

lukaszcl:
The flaky ratio is now displayed with up to four decimal places when it is very small, such as 0.0687%. Example workflow run: https://github.com/smartcontractkit/chainlink/actions/runs/12667914659


AKhozya (approved):

Tofel (approved):
although I'd add a test where flakyness ration is sth like `0.0008%` to see that latest change in action

lukaszcl:
> although I'd add a test where flakyness ration is sth like `0.0008%` to see that latest change in action

@Tofel see "Mixed skipped and executed tests" test

PR #1545, opened 2025-01-09
perf tool enhancement

Tofel:
Direct query comparison:
* add threshold validation, 
* nil report validation, 
* better readability for infinite metric change (`999%` instead of `100%` since in reality the change is infinite)
* couple more unit tests for edge cases

---

## Why
The changes refine the calculation of percentage differences, especially for edge cases where previous values are zero, and enhance the validation and comparison logic in the benchmark spy reporting functionality. This includes handling cases with nil reports, incorrect thresholds, and transitions from zero to non-zero values more gracefully, which improves the accuracy and reliability of performance comparison over time.

## What
- **wasp/benchspy/report.go**
  - Modified `calculateDiffPercentage` function to handle edge cases, especially when the previous value is 0, by introducing conventions for infinite changes and complete improvements.
  - Added checks in `CompareDirectWithThresholds` to immediately return errors if one or both reports are nil, ensuring the function handles uninitialized reports.
  - Introduced `validateThresholds` function to ensure that the thresholds for median, P95, max latency, and error rate are within the acceptable range of 0 to 100.
- **wasp/benchspy/report_test.go**
  - Enhanced test cases to cover new scenarios, including handling of nil reports, incorrect threshold values, zero to non-zero transitions, and non-zero to zero transitions. This ensures the new logic behaves as expected under a wide range of conditions.

### DIFF
--- a/wasp/benchspy/report.go
+++ b/wasp/benchspy/report.go
@@ -3,6 +3,7 @@ package benchspy
 import (
 	"context"
 	"encoding/json"
+	goerrors "errors"
 	"fmt"
 	"os"
 	"strings"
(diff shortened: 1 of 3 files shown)


Tofel:
code duplication content is related to unit tests, not gonna fix it 🤷

wasp/benchspy/report.go
```diff
@@ -150,21 +150,29 @@ func MustAllPrometheusResults(sr *StandardReport) map[string]model.Value {
...
-	return diffPrecentage
+	return (current - previous) / previous * 100
 }
 
 // CompareDirectWithThresholds evaluates the current and previous reports against specified thresholds.
 // It checks for significant differences in metrics and returns any discrepancies found, aiding in performance analysis.
 func CompareDirectWithThresholds(medianThreshold, p95Threshold, maxThreshold, errorRateThreshold float64, currentReport, previousReport *StandardReport) (bool, map[string][]error) {
+	if currentReport == nil || previousReport == nil {
+		return true, map[string][]error{
+			"initialization": {errors.New("one or both reports are nil")},
```
skudasov, line 172: Maybe we can use a struct as an error to make this API more discoverable/typed? Can be done in other PRs but looks weird

skudasov (approved):

lukaszcl (approved):

PR #1590, opened 2025-01-21
Define resources for common components: NodeSet, Blockchain, PostgreSQL

skudasov:
---

## Why
The changes introduce a new section in the documentation for Components Resources and implement resource limits for containers in various blockchain components, Chainlink nodes, and PostgreSQL components. This ensures efficient resource usage and prevents any single component from monopolizing system resources, improving overall system stability and performance.

## What
- Added `Components Resources` section to the documentation.
- Set cgroups resource for containers in the changeset `v0.4.7`.
- Added resource limits for containers in blockchain components (`anvil.go`, `aptos.go`, `besu.go`, `geth.go`, `solana.go`, `sui.go`, `tron.go`), Chainlink nodes (`clnode.go`), and PostgreSQL components (`postgres.go`).
- Introduced `ContainerResources` structure to define CPU and memory limits for containers.
- Implemented `ResourceLimitsFunc` to apply resource limits based on `ContainerResources`.
- Added resource limits configuration in examples (`smoke.toml`, `smoke_limited_resources.toml`).

### DIFF
--- a/book/src/framework/components/resources.md
+++ b/book/src/framework/components/resources.md
@@ -1,6 +1,7 @@
 # Components Resources
 
 You can use `resources` to limit containers CPU/Memory for `NodeSet`, `Blockchain` and `PostgreSQL` components.
+
 ```toml
 [blockchain_a.resources]
 cpus = 0.5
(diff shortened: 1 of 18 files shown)


book/src/framework/components/resources.md
```diff
@@ -0,0 +1,24 @@
...
+[nodeset.node_specs.node.resources]
+cpus = 1
+memory_mb = 1048
+```
+
+Read more about resource constraints [here](https://docs.docker.com/engine/containers/resource_constraints/).
+
+We are using `cpu-period` and `cpu-quota` for simplicity, and because it's working with an arbitrary amount of containers, it is absolute.
+
+Memory swapping is off if you specify `resources` key.
```
Tofel, line 22: can you give an example? it's not clear to me.

book/src/framework/components/resources.md
```diff
@@ -0,0 +1,24 @@
...
+memory_mb = 2048
+
+[nodeset.node_specs.node.resources]
+cpus = 1
+memory_mb = 1048
+```
+
+Read more about resource constraints [here](https://docs.docker.com/engine/containers/resource_constraints/).
+
+We are using `cpu-period` and `cpu-quota` for simplicity, and because it's working with an arbitrary amount of containers, it is absolute.
```
Tofel, line 20: could you elaborate? how does `cpu = 0.5` translate into `cpu-period` and `cpu-quota`? we have 1 setting that is then converted into 2?

framework/docker.go
```diff
@@ -320,3 +320,28 @@ func ExecContainer(containerName string, command []string) (string, error) {
 	L.Info().Str("Output", string(output)).Msg("Command output")
 	return string(output), nil
 }
+
+type ContainerResources struct {
+	CPUs     float64 `toml:"cpus"`
+	MemoryMb int     `toml:"memory_mb"`
```
Tofel, line 326: why not `uint`?

Tofel (approved):

PR #1595, opened 2025-01-23
Parrot Server

kalverra:
Adds a new, generalized and highly configurable Parrotserver to replace our current Mockserver/Killgrave mishmash.

* Is super simple to use
* Can be used locally and remotely
* Has ability to easily record all calls made to it
* AMD64 and ARM64 compatible


parrot/recorder.go
```diff
@@ -0,0 +1,142 @@
...
+	Body       []byte      `json:"body"`
+}
+
+func NewRecorder() (*Recorder, error) {
+	r := &Recorder{
+		recordChan: make(chan *RouteCall),
+		errChan:    make(chan error),
+	}
+
+	// TODO: Will need a way to send out the URL to an external service (e.g. Parrotserver running in a docker container)
```
skudasov, line 49: Here is our guide for components: https://smartcontractkit.github.io/chainlink-testing-framework/developing/developing_components.html
It can solve this problem, return it as an `*Output* as other components do, and, ideally, move it under "components" dir, wdyt?

parrot/recorder_test.go
```diff
@@ -0,0 +1,137 @@
...
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestResponseWriterRecorder(t *testing.T) {
```
skudasov, line 13: I think it's worth adding a test case for parallel usage, in current version in `CTFv2` we have a mutex in our recorder

parrot/examples_test.go
```diff
@@ -0,0 +1,191 @@
...
+
+	fmt.Println(resp.StatusCode)
+	body, _ := io.ReadAll(resp.Body)
+	fmt.Println(string(body))
+	// Output:
+	// 200
+	// Squawk
+}
+
+func ExampleRoute() {
```
skudasov, line 56: Maybe let's do an example with NodeSet and some blockchain to show how to mock EAs? Can be done later, though
https://github.com/smartcontractkit/chainlink-testing-framework/blob/main/framework/examples/myproject/smoke_test.go#L24

skudasov (changes requested):

parrot/recorder.go
kalverra: That works fine for me. Will look at moving it.

parrot/recorder_test.go
kalverra: Agreed, adding

parrot/examples_test.go
kalverra: Agreed it would be a helpful example, but I think it's a bit out of scope for just this tool. We should include examples like this as part of a larger, more comprehensive setup.

gheorghestrimtu (approved):

lukaszcl (approved):
lgtm. Good job @kalverra!

skudasov (approved):

PR #1757, opened 2025-04-04
remove interactive mode, add aptos cmd and custom port bindings, expose containers rm command

skudasov:


framework/components/blockchain/aptos.go
```diff
@@ -21,9 +21,8 @@ func defaultAptos(in *Input) {
 	if in.Image == "" {
 		in.Image = "aptoslabs/tools:aptos-node-v1.27.2"
 	}
-	if in.Port != "" {
-		framework.L.Warn().Msg("'port' field is set but only default port can be used: 8080")
-	}
+	framework.L.Warn().Msg("Aptos node API can only be exposed on port 8080!")
```
Tofel, line 24: curious: why?

framework/docker.go
```diff
@@ -394,3 +416,40 @@ func ResourceLimitsFunc(h *container.HostConfig, resources *ContainerResources)
...
+
+func GenerateCustomPortsData(portsProvided []string) ([]string, nat.PortMap, error) {
+	exposedPorts := make([]string, 0)
+	portBindings := nat.PortMap{}
+	customPorts := make([]string, 0)
+	for _, p := range portsProvided {
+		if strings.Contains(p, ":") {
+			pp := strings.Split(p, ":")
+			if len(pp) != 2 {
+				return nil, nil, errors.New("custom_ports has ':' but you must provide both ports")
```
Tofel, line 428: nit: maybe also attach the value of `p` to the error message?

Tofel (approved):

framework/components/blockchain/aptos.go
skudasov: Not exposed in their CLI, can't override that port

PR #1774, opened 2025-04-09
JD fork test example

skudasov:


framework/cmd/observability/compose/conf/provisioning/dashboards/workflow-engine/engine.json
```diff
@@ -258,13 +258,54 @@
...
+        "sortOrder": "Descending",
+        "wrapLogMessage": true
+      },
+      "targets": [
+        {
+          "datasource": {
+            "type": "loki"
+          },
+          "editorMode": "code",
+          "expr": "{service_name=\"unknown_service:chainlink\"} !~ \"heartbeat\"\n| json | __error__ = \"\"",
```
Tofel, line 291: we didn't manage to fix that?

framework/docker.go
```diff
@@ -112,6 +112,51 @@ func NewDockerClient() (*DockerClient, error) {
 	return &DockerClient{cli: cli}, nil
 }
 
+// ExecContainer executes a command inside a running container by name and returns the combined stdout/stderr.
+func (dc *DockerClient) ExecContainer(containerName string, command []string) (string, error) {
+	L.Info().Strs("Command", command).Str("ContainerName", containerName).Msg("Executing command")
+	ctx := context.Background()
```
Tofel, line 118: how about using ctx with some timeout?

Tofel (approved):

framework/cmd/observability/compose/conf/provisioning/dashboards/workflow-engine/engine.json
skudasov: Let's update it to .*chainlink, labels can change and we don't want to reflect these changes on the dashboard

framework/docker.go
skudasov: It doesn't make sense to me right now since our tests do not use context everywhere, so they'll provide context.Background() but put it as a ticket so we can add it later for all our utilities https://smartcontract-it.atlassian.net/browse/DX-481

PR #1829, opened 2025-05-13
Solidity layout helper for anvil_setStorageAt

skudasov:


framework/.changeset/v0.7.8.md
```diff
@@ -0,0 +1 @@
+- Solidity storage layout helper for anvil_setStorageAt + tests + guide
```
Tofel, line 1: please bump to 0.7.9, because 0.7.8 will be released very shortly (or would have already been released by the time you read it)

framework/evm_storage/layout_api_test.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+
+	slot := layout.ArraySlot("a_signers", 1)
+	data := encodeFunc("0x00000000000000000000000000000000000000a5", 255, 42)
+	fmt.Printf("setting slot: %s with data: %s\n", slot, data)
+	r := rpc.New(testRPCURL, nil)
+	err = r.AnvilSetStorageAt([]interface{}{contractAddr, slot, data})
+	require.NoError(t, err)
+	// verify it manually
+	// cast call 0x5FbDB2315678afecb367f032d93F642f64180aa3 "getASigner(uint256)(address,uint8,uint8)" --rpc-url http://localhost:8545 1
```
Tofel, line 51: any reason why we can't verify automatically?

Tofel (approved):

framework/evm_storage/layout_api_test.go
skudasov: We are already doing that in a second test file, This test is more like a playground and API example

PR #1849, opened 2025-05-23
Add Aptos CLI and LOOP plugin config to CITestConf

joaoluisam:


kdihalas (approved):

tools/citool/cmd/types.go
```diff
@@ -21,6 +21,8 @@ type CITestConf struct {
 	RemoteRunnerMemory         string            `yaml:"remote_runner_memory" json:"remote_runner_memory"`
 	PyroscopeEnv               string            `yaml:"pyroscope_env" json:"pyroscope_env"`
 	Triggers                   []string          `yaml:"triggers" json:"triggers"`
+	InstallPluginsPublic       bool              `yaml:"install_plugins_public" json:"install_plugins_public"`
```
Tofel, line 24: are these plugins Apots-specific? or general? if the former pls adjust naming, if latter, maybe add a comment that explains what plugins we are talking about?

PR #1938, opened 2025-06-25
[DX-1118] local beholder support

Tofel:
chip ingress + red panda
+ functions to register topics and protos

All in separate go module to limit blast radius of all dependencies required by docker-compose wrapped in `testcontainers-go`


framework/components/dockercompose/chip_ingress_set/chip_ingress.go
```diff
@@ -0,0 +1,261 @@
...
+	}
+	defer cli.Close()
+
+	timeout := time.After(1 * time.Minute)
+	tick := time.Tick(500 * time.Millisecond)
+
+	// so let's try to connect to a Docker network a couple of times, there must be a race condition in Docker
+	// and even when network sandbox has been created and container is running, this call can still fail
+	// retrying is simpler than trying to figure out how to correctly wait for the network sandbox to be ready
+	var connectNetwork = func(networkName string) error {
```
skudasov, line 132: Can't we just be on the same network? Seems like we either need to remove the default network from CTF or to configure this compose to use the same CTF network

Tofel: and what's the problem with this approach?

skudasov (approved):

PR #2516, opened 2026-04-09
[DX-3763] dump inuse and memory profile

Tofel:


framework/components/clnode/default.go
```diff
@@ -19,11 +19,6 @@ const defaultConfigTmpl = `
 [Log]
 Level = 'debug'
 
-[Pyroscope]
-ServerAddress = 'http://pyroscope:4040'
-Environment = 'local'
-LinkTracesToProfiles = true
```
Tofel, line 25: When Pyroscope is configured, then dumping admin profiles fails with `error collecting /v2/debug/pprof/profile?seconds=1: status 500: Could not enable CPU profiling: cpu profiling already in use`

skudasov (dismissed):

PR #2522, opened 2026-04-22
Add linking service and Vault JWT issuer framework components

prashantkumar1982:
## Summary
- Add a generic `framework/components/linkingservice` component with a Docker image, gRPC server, and runtime helpers so tests can use a first-class, reusable linking service managed by the framework instead of repo-local ad hoc mocks.
- Add a `framework/components/vaultjwtissuer` component with a Docker image, JWKS/OpenID HTTP server, and JWT helper APIs so Vault-style JWT flows can run against a Dockerized support service that is owned by the framework lifecycle.
- Add the module and dependency updates required by those new components, including the linking-service proto/gRPC/protobuf wiring and newer security-patched dependencies. These changes were needed both to compile the new components and to satisfy dependency-review on the new dependency graph.
- Fix component image build contexts so the new service images build correctly from CI and downstream repos. This was needed because the initial component packaging did not match the build context layout expected by the framework.
- Update CI and dev tooling to handle the newer module graph, including letting the framework lint job use a `golangci-lint` binary built with a Go version compatible with the framework module. This was needed because the framework module now resolves to `go 1.25.0`, which the old lint binary could not target.
- Harden existing CI jobs that became flaky while validating this PR by switching JD and chip-ingress Postgres images away from Docker Hub rate-limit/auth sensitivity, extending chip-ingress startup deadlines, adding a short Seth Geth startup retry, and making the scalability example fail fast on generator errors instead of hanging until the outer test timeout. These changes were needed to keep unrelated framework checks reliable while the new components were being introduced.


framework/components/dockercompose/chip_ingress_set/chip_ingress.go
```diff
@@ -66,6 +66,7 @@ func defaultChipIngress(in *Input) *Input {
 
 const (
 	DEFAULT_STACK_NAME = "chip-ingress"
+	stackStartupTimeout = 5 * time.Minute
```
Tofel: why do we need such long startup timeout?

framework/components/linkingservice/linkingservice.go
```diff
@@ -0,0 +1,170 @@
...
+
+	"github.com/docker/docker/api/types/container"
+	"github.com/docker/go-connections/nat"
+	"github.com/google/uuid"
+	tc "github.com/testcontainers/testcontainers-go"
+	tcwait "github.com/testcontainers/testcontainers-go/wait"
+)
+
+const (
+	DefaultImage         = "local-cre-linking-service:latest"
```
Tofel: let's avoid using mutable tags as that has bitten us more than once, which in this case means we can't have a default image or that we pin it to an immutable tag

framework/components/linkingservice/linkingservice.go
```diff
@@ -0,0 +1,170 @@
...
+}
+
+const defaultNetworkName = "ctf"
+
+func defaultTCLabels() map[string]string {
+	return map[string]string{
+		"framework": "ctf",
+		"logging":   "promtail",
+	}
+}
```
Tofel: we already have a function for that in `framework/docker.go` called `DefaultTCLabels()`

framework/components/linkingservice/linkingservice.go
```diff
@@ -0,0 +1,170 @@
...
+		return err
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("linking service health request failed with status %s", resp.Status)
+	}
+	return nil
+}
+
+const defaultNetworkName = "ctf"
```
Tofel: we have a constant for that in `framework/config.go`, unless you want to avoid depending on framework?

framework/components/vaultjwtissuer/vaultjwtissuer.go
```diff
@@ -0,0 +1,143 @@
...
+
+	"github.com/docker/docker/api/types/container"
+	"github.com/docker/go-connections/nat"
+	"github.com/google/uuid"
+	tc "github.com/testcontainers/testcontainers-go"
+	tcwait "github.com/testcontainers/testcontainers-go/wait"
+)
+
+const (
+	DefaultImage         = "local-cre-vault-jwt-issuer:latest"
```
Tofel: same, let's not use mutable tags

framework/components/vaultjwtissuer/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+
+COPY framework /src/framework
+COPY wasp /src/wasp
+RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -trimpath -ldflags="-s -w" -o /out/vault-jwt-issuer ./components/vaultjwtissuer/cmd/vault-jwt-issuer
+
+FROM gcr.io/distroless/base-debian12
+
+COPY --from=build /out/vault-jwt-issuer /vault-jwt-issuer
+EXPOSE 18123
+ENTRYPOINT ["/vault-jwt-issuer"]
```
Tofel, line 17: would be great to have a single smoke test for each component here on the CTF side similar to `framework/examples/chip_ingress/smoke_chip_ingress_test.go`

framework/components/dockercompose/chip_ingress_set/chip_ingress.go
prashantkumar1982: reverted back to 2 mins

framework/components/vaultjwtissuer/Dockerfile
prashantkumar1982: Added:
linkingservice_smoke_test.go (line 18)
vaultjwtissuer_smoke_test.go (line 16)

.github/workflows/linters.yml
```diff
@@ -32,9 +32,22 @@ jobs:
         uses: actions/setup-go@v5
         with:
           go-version-file: ${{ matrix.dir }}/go.mod
+      - name: Install framework golangci-lint with module Go version
+        if: matrix.dir == 'framework'
+        run: |
+          mkdir -p "$RUNNER_TEMP/bin"
+          GOBIN="$RUNNER_TEMP/bin" go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.64.5
+          echo "$RUNNER_TEMP/bin" >> "$GITHUB_PATH"
```
Tofel, line 40: curious: why is this needed?

framework/components/linkingservice/linkingservice.go
```diff
@@ -0,0 +1,154 @@
...
+	if in.AdminPort == 0 {
+		in.AdminPort = DefaultAdminPort
+	}
+	ApplyImageOverride(in)
+	if in.ContainerName == "" {
+		in.ContainerName = DefaultContainerName
+	}
+}
+
+func ApplyImageOverride(in *Input) string {
```
Tofel, line 54: why does it need to be public? I think it makes more sense to have it private and run it inside `NewWithContext`, so that callers, like `chainlink`/local CRE do not need to be aware that they should run this override func

framework/components/vaultjwtissuer/client.go
```diff
@@ -0,0 +1,169 @@
...
+		claims.Issuer = c.LocalIssuerURL()
+	}
+	if claims.Audience == "" {
+		claims.Audience = DefaultAudience
+	}
+
+	return signToken(c.privateKey, claims)
+}
+
+func NormalizeIssuerURL(raw string) string {
```
Tofel, line 94: why does it have to be public?

skudasov:
Great, do you need any help @prashantkumar1982? Do you think we can finish this PR, or have priorities changed?

PR #2780, opened 2026-09-01
chore: implement phase 2

Tofel:
Fixes retry-error conflation, measures full poll latency, and hardens the Source test doubles for concurrent use.

Review focus: `source.go` (error vs retry classification, latency timing) and `source_fake_test.go` (concurrency safety).


grafana-alertcheck/internal/gate/source.go
```diff
@@ -0,0 +1,373 @@
...
+	"net/url"
+	"strconv"
+	"strings"
+	"time"
+)
+
+// skewHardLimit is one of §5's filled-in values (basis: §16; §22.11 asserts
+// 120s errors, 30s does not). It belongs in schedule.go's named-constants
+// block once P4 exists; defined here because P2 needs it first.
+const skewHardLimit = 60 * time.Second
```
Tofel, line 20: Clock skew is only ever needed to place the runner-defined window edges `[from, to+grace]` onto Grafana's own timeline, so it's a single-point reconciliation, not a factor in any duration-based check. The 60s hard limit is therefore a sanity gate, not a correction: skew is estimated as `serverDate − (t_send+t_recv)/2 with error RTT/2`, and that estimation is only trusted while the two clocks are in rough agreement. Past 60s the offset is treated as a broken or misconfigured clock, so the gate fails closed rather than align fine-grained window work against a clock it no longer trusts.

grafana-alertcheck/internal/gate/source.go
```diff
@@ -0,0 +1,373 @@
...
+		*field = n
+	}
+	return v, nil
+}
+
+// supportedGrafanaMin and supportedGrafanaMax bound the platform this gate is
+// verified against (§2.7 control 2, §21.5): >= 13.0.0, < 14.0.0.
+var (
+	supportedGrafanaMin = grafanaVersion{13, 0, 0}
+	supportedGrafanaMax = grafanaVersion{14, 0, 0} // exclusive
```
Tofel, line 140: better safe than sorry, we don't want to risk subtle behaviour changes.

grafana-alertcheck/internal/gate/source.go
```diff
@@ -0,0 +1,373 @@
...
+		return requestResult{}, &TransportError{Err: doErr}
+	}
+	defer resp.Body.Close()
+
+	b, readErr := io.ReadAll(resp.Body)
+	tBodyRead := s.clock.Now()
+	if readErr != nil {
+		return requestResult{}, &TransportError{Err: fmt.Errorf("read response body (status %d): %w", resp.StatusCode, readErr)}
+	}
+	latency := tBodyRead.Sub(tSend)
```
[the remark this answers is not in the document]
Tofel: good catch

PR #2804, opened 2026-09-04
chore: remove bitnami helm repo registration (IP-3481)

tvc-jeffwelling:
**What:** Removes the three Bitnami Helm repo registrations from CTF (IP-3481):

- `lib/k8s/Dockerfile.base` - drop `helm repo add bitnami https://charts.bitnami.com/bitnami`
- `lib/k8s/examples/link/Dockerfile` - same line
- `.helm-repositories.yaml` - remove the `bitnami` entry (file is an orphan from the removed nix shell feature, TT-1042; nothing references it)

No chart pulls by repo name `bitnami` exist anywhere in the repo - these are pure registrations. Kafka/influx charts come via `BITNAMI_PRIVATE_REGISTRY` (separate scan rows, separate PR).

**Why this commit uses `--no-verify`:** the repo's pre-commit hookset is broken on main for everyone:

- `go-lint` fails: `just lint-all` calls `just lint havoc` (justfile:25) but no top-level `havoc/` dir exists on main.
- `go-mod-tidy` fails on any dirty tree, and `go-lint`'s `golangci-lint --fix` (justfile:17) mutates files, which then fails the tidy diff - the two hooks contradict each other.
- `typos` fails on ~35 pre-existing typos across the repo (none in the files changed here; a few are false positives in vendored code, one is inside a private-key literal).

None of the skipped checks involve files in this change. Verified via **probe PR #2803** (empty commit): CI runs `just lint` per module + full test matrix + workflow checks - **all green**. CI does not run pre-commit.

### DIFF
--- a/.helm-repositories.yaml
+++ b/.helm-repositories.yaml
@@ -5,15 +5,6 @@ repositories:
     certFile: ''
     insecure_skip_tls_verify: false
     keyFile: ''
-    name: bitnami
-    pass_credentials_all: false
-    password: ''
-    url: https://charts.bitnami.com/bitnami
-    username: ''
-  - caFile: ''
-    certFile: ''
-    insecure_skip_tls_verify: false
-    keyFile: ''
     name: chainlink-qa
     pass_credentials_all: false
     password: ''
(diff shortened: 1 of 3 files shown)


chainchad (approved):
LGTM but @skudasov should confirm

skudasov (approved):

